PR #6, opened 2018-07-25
Fix naming of types for fields nested under unions

lemonmade:
Our type generation had a weird edge case that breaks: when you have a union type, and multiple members of that union each have a field whose name matches between them, we generate a non-unique interface name. This PR namespaces all fields under unions in order to avoid such conflicts (see the added test for details).

cc/ @tsov since this was uncovered by online store [here](https://github.com/Shopify/online-store-web/pull/586/files).


tsov:
🎉

PR #7, opened 2018-07-29
Support `any` type for custom scalars

nikhilmat:
Hi there!

Thanks for writing a great library! This has been immensely useful for us.

In our GraphQL backend, we have a custom scalar used for representing JSON. Currently, this always get represented as a `string` type when the Typescript types are generated. For our custom scalar case, we want this to be an `any` type, so that it can be used like object in our Typescript code.

This PR adds a CLI option to specify the type for custom scalars as any - I didn't see a way for this to be configured on a per scalar basis, but open to any feedback if there's a better way to implement this!

### DIFF
--- a/packages/graphql-typescript-definitions/src/cli.ts
+++ b/packages/graphql-typescript-definitions/src/cli.ts
@@ -33,6 +33,12 @@ const argv = yargs
     type: 'boolean',
     describe: 'Add a __typename field to every object type',
   })
+  .option('scalar-any', {
+    required: false,
+    default: false,
+    type: 'boolean',
+    describe: 'Generate custom scalars as `any` type',
+  })
   .option('enum-format', {
     required: false,
     type: 'string',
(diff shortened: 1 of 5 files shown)


nikhilmat:
@cartogram is this a PR you would accept? Wondering if there's another way folks are handling custom scalars that are not strings.

nikhilmat:
@lemonmade is this a PR you would accept? Wondering if there's another way folks are handling custom scalars that are not strings.

lemonmade:
Glad it’s been helpful! I am definitely open to something along these lines. They way I was thinking about this is that you would specify a custom type on a per-scalar basis, so you could customize to your heart’s content. That would be a bit more involved as you’d need to basically pass JSON to do it in the CLI. Thoughts?

Out of curiosity, what are you using as a GraphQL client that deserializes the data such that a JSON scalar come stay out as a real object? As far as I know, Apollo does not have a native way of doing this.

lemonmade:
@nikhilmat I just added a change to allow passing custom GraphQL types for custom scalars, look for it in the next version!

nikhilmat:
@lemonmade amazing! i just got started using it today, works great :)

PR #11, opened 2018-08-14
Add `showPasses` argument

cartogram:
Part of https://github.com/Shopify/sewing-kit/issues/820

Adds a flag to conditionally show the `Pass`/valid fixtures, otherwise, do not output those.


cartogram:
@GoodForOneFare updated with tests ✌️

PR #13, opened 2018-08-21
Adds check for filename mismatch

cartogram:
Part of https://github.com/Shopify/sewing-kit/issues/811

This PR adds a check to see if the query is named the same as the file that contains it and emits an error if not.

This PR does not add a check for graphql files that contain Fragments. I will provide a similar PR for fragments following this one.


packages/graphql-typescript-definitions/test/cli.test.ts
```diff
@@ -27,6 +27,12 @@ describe('cli', () => {
       await execDetails(cliCommandForFixtureDirectory('missing-types')),
     ).toMatchSnapshot();
   });
+
+  it('fails when the file has different name than the query defined within it', async () => {
+    expect(
+      await execDetails(cliCommandForFixtureDirectory('misnamed-file')),
+    ).toMatchSnapshot();
```
ismail-syed: [We do not use Jest snapshot tests](https://github.com/Shopify/web-foundation/blob/master/Decision%20records/03%20-%20We%20do%20not%20use%20Jest%20snapshot%20tests.md) but we still do. 
IMO, we should try to stop the bleeding from getting worse. Can we refactor this test to not use snapshots. We can revisit the already existing ones later.

GoodForOneFare: I'm okay with this if an issue to refactor is raised.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -278,3 +296,31 @@ function groupOperationsAndFragmentsByFile({
...
+
+function getFileNameFromPath(path: string) {
+  const fileName = basename(path);
+  const fileExtension = extname(path);
+  const extensionLength = fileExtension.length;
+
+  return fileName.slice(0, -extensionLength);
+}
+
+function checkFileNameMismatch(file: File) {
```
ismail-syed: I'm finding this a bit difficult to follow. Something more like `getFileNameFromPath` is easier to read.

```
const {operation} = file;

if (!operation) {
  return false;
}

const {filePath, operationName, operationType} = operation;
const fileName = getFileNameFromPath(operation.filePath);
const type = toCapitalized(operationType);

return fileName !== `${operationName}${type}`;
```

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -278,3 +296,31 @@ function groupOperationsAndFragmentsByFile({
...
+
+  return (
+    operation &&
+    getFileNameFromPath(operation.filePath) !==
+      `${operation.operationName}${toCapitalized(operation.operationType)}`
+  );
+}
+
+function getExpectedFileNameFromOperation(operation: Operation) {
+  return `${operation.operationName}${toCapitalized(
```
ismail-syed: Same here. I think this could use a small refactor to make it more readable.

packages/graphql-typescript-definitions/src/index.ts
ismail-syed: 😆 Don't know if that snippet even works. But I think you get what I mean

cartogram: Agreed 👍

GoodForOneFare:
Doing a 🎩of this in the u2 repo, and it doesn't fix #811's scenario.  Stepping into the debugger, I see that:
* u2 had two queries with the same name
* `graphql-typescript-definitions` smooshes the schema + queries into a single AST: https://github.com/Shopify/graphql-tools-web/blob/a327937c63f0575260ee30f75bf07626aac3b141/packages/graphql-typescript-definitions/src/index.ts#L161-L165
* `ast.operations` is a map of `operation name => Operation`: https://github.com/Shopify/graphql-tools-web/blob/a327937c63f0575260ee30f75bf07626aac3b141/packages/graphql-tool-utilities/ast.ts#L75-L80
* So if two queries have the same name, one overwrites the other in `ast.operations` 😬 
* Meaning that a single `.d.ts` is generated, when we'd expect two

-----

### How can I reproduce this?
1. Create a `grapql-typescript-definitions` fixture with:
    * `foo/query1.graphql` containing `query MyQuery { foo }`
    * `bar/query2.graphql` containing `query MyQuery { bar }`
2. Generate type definitions
3. Note that only one `.d.ts` file is generated 😭

cartogram:
@GoodForOneFare do you feel this PR is still relevant and wrapped up _or_ better to revisit as a lint thing and closed?

GoodForOneFare:
I think revisting as a lint thing.  It's a little too opinionated for a type generation lib.

PR #15, opened 2018-08-23
Error on duplicate query names

cartogram:
Branches from https://github.com/Shopify/graphql-tools-web/pull/13#issuecomment-415279569
Part of Shopify/sewing-kit#811

Adds an error for when multiple queries share the same name. 

**Question:** Should we crash or just log the error?

```shell
ERROR  Graphql operations must have unique names. Found 2 or more operations with the name "FindPerson"
```

----

### 🎩  Instructions for `web`

(may vary depeding on where you store project locally)

1. run `yarn build` from the root of this repo.
2. copy the built source to web `cp -R packages/graphql-typescript-definitions/lib/* ~/src/github.com/Shopify/web/node_modules/\@shopify/sewing-kit/node_modules/graphql-typescript-definitions/lib/
`
----
Screenshot for reference

screen shot 2018-08-24 at 1 00 24 pm

### DIFF
--- a/packages/graphql-typescript-definitions/src/index.ts
+++ b/packages/graphql-typescript-definitions/src/index.ts
@@ -3,6 +3,8 @@ import {
   buildClientSchema,
   GraphQLSchema,
   DocumentNode,
+  DefinitionNode,
+  OperationDefinitionNode,
   parse,
   Source,
   concatAST,
@@ -158,6 +160,17 @@ export class Builder extends EventEmitter {
     this.emit('start');
     let ast: AST;
 
+    const duplicateNames = getDuplicateOperationNames(this.documentCache);
+
+    if (duplicateNames.length) {
+      duplicateNames.forEach((name) => {
+        const error = new Error(
+          `Graphql operations must have unique names. Found 2 or more operations with the name "${name}"`,
+        );
+        this.emit('error', error);
+      });
+    }
+
     try {
       ast = compile(
         this.schema,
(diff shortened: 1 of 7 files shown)


packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -158,6 +160,17 @@ export class Builder extends EventEmitter {
     this.emit('start');
     let ast: AST;
 
+    const duplicateNames = getDuplicateOperationNames(this.documentCache);
+
+    if (duplicateNames.length) {
+      duplicateNames.forEach((name) => {
+        const error = new Error(
+          `Graphql operations must have unique names. Found 2 or more operations with the name "${name}"`,
```
cartogram: missing a period

GoodForOneFare: If we have file info available, we shouldn't make users hunt down specific files.  An error like this provides good context for self-service:

neutron zsh 2018-08-26 14-28-03


Scrappy implementation is inside the details block:
<details>
```ts
  const duplicateOperations = getDuplicateOperations(this.documentCache);
  if (duplicateOperations.length) {
    duplicateOperations.forEach(({name, files}) => {
      const error = new Error(
        `Graphql operations must have a unique name. The operation "${name}" is declared in:\n    ${files.join(
          '\n    ',
        )}`,
      );
      this.emit('error', error);
    });
  }

function getDuplicateOperations(documents: Map<string, DocumentNode>) {
  const operations = new Map<string, Set<string>>();

  Array.from(documents.entries()).forEach(([path, document]) => {
    document.definitions.filter(isOperation).forEach((definition) => {
      const {name} = definition as OperationDefinitionNode;
      if (name && name.value) {
        const map = operations.get(name.value);
        if (map) {
          map.add(path);
        } else {
          operations.set(name.value, new Set([path]));
        }
      }
    });
  });

  return Array.from(operations.entries())
    .filter(([, files]) => files.size > 1)
    .map(([operation, files]) => {
      return {name: operation, files: Array.from(files)};
    });
}
```
</details>

GoodForOneFare: In watch mode, we shouldn't generate `.d.ts` files for clashing files.

PR #19, opened 2018-08-28
adding support for multiple schemas

patsissons:
Prior to this PR, this tool was only capable of processing a single schema within a workspace. The [Prisma `.graphqlconfig` (`graphql-config`)](https://github.com/prisma/graphql-config) format is gaining community support (see the [VSCode extension](https://github.com/prisma/vscode-graphql)) and is much more expressive, specifically in handling multiple schemas. This PR is mostly a drop in replacement, only requiring a small migration from the old `.gqlconfig` to the new `.graphqlconfig` format on consumer projects.

The intergration of `graphql-config` essentially builds a workspace `GraphQLConfig` and queries this instance whenever we want to process a graphql file (query, mutation, fragment, or schema). The top level config instance holds a collection of projects, each project is linked to a schema and includes/excludes glob lists. Project configs also support an `extensions` _sandbox_ element that we can optionally use to configure a custom schema types output path (see `schemaTypesPath` below). Our workspace config can tell us which project a file matches with (though glob intersection can cause problems, see *caveats* section below).

Any included files (typically these would be `*.graphql` are processed into `*.graphql.d.ts`. Schema files are only processed into schema types (either `types.ts` for nameless projects, or `${projectName}-types.ts` for named projects) but this can be overridden by adding a `schemaTypesPath` extension with the path to use instead which allows customization over the name of the schema types file (see below for an example). The `graphql-config` tooling supports linking a project to either `schema.json` or `schema.graphql`.

CLI output has been cleaned up a little as well, primarily to produce consistent and deterministic output. This is achieved by buffering outputs from `build` events (first clearing buffers on `start` events) and printing them in order on `end` events.

the `dom` lib has been added to the `tsconfig.json`. Normally, this is not suggested for a node project, however in this case our `graphql-config` dependency (and by extension, its dependency `graphql-request`) expects the `dom` lib. Additionally, Sewing Kit [also uses dom](https://github.com/Shopify/sewing-kit/blob/master/tsconfig.json#L8) so we don't have much of a choice in the matter here.

### ⚠️ Caveats

* glob intersections can cause unpredictable behaviour. The first project that matches a file will be used for schema linking, which depends on the `GraphQLConfig.getProjects` function, which is iterating the configured projects. This could mean that naming of projects in `.graphqlconfig` could have an effect on runtime execution. A workaround to prevent this is to use globs that are namespaced enough to prevent overlap or add excludes to projects with wide reaching include globs.
* [globbing does not support absolute paths](https://github.com/stephen/vscode-graphql/issues/23). the globbing in `graphql-config` will reduce files to match to relative paths against the `.graphqlconfig` file, and the globbing tool (`minimatch`) cannot match a relative file against an absolute glob pattern. The workaround to prevent this is to simply use relative globs.

### 🎩  Instructions for `sewing-kit`

`yarn --cwd packages/graphql-typescript-definitions build && cp -R packages/graphql-typescript-definitions/lib/* ~/src/github.com/Shopify/sewing-kit/node_modules/graphql-typescript-definitions/lib/`

### 🎩  Instructions for `web`

(may vary depeding on where you store project locally)

`yarn --cwd packages/graphql-typescript-definitions build && cp -R packages/graphql-typescript-definitions/lib/* ~/src/github.com/Shopify/web/node_modules/\@shopify/sewing-kit/node_modules/graphql-typescript-definitions/lib/`

You may want to also install the [Prisma VSCode Extension](https://marketplace.visualstudio.com/items?itemName=Prisma.vscode-graphql) to see the tooling provide hover text, auto-complete, and intellisense to `*.graphql` files.

I have 🎩'd this with `web` with the `.graphqlconfig` below and everything worked as expected. The custom schema does not get validated against the core schema. The custom schema types file is generated (`{projectName}-types.ts`).

```json
{
  "projects": {
    "bourgeois": {
      "schemaPath": "app/sections/Capital/graphql/bourgeois/schema.json",
      "includes": ["app/sections/Capital/**/graphql/bourgeois/*.graphql"]
    },
    "core": {
      "schemaPath": "build/schema.json",
      "includes": ["app/**/*.graphql"],
      "excludes": ["app/**/graphql/bourgeois/*.graphql"],
      "extensions": {
        "schemaTypesPath": "app/types/graphql.ts"
      }
    }
  }
}
```

### DIFF
--- a/packages/graphql-typescript-definitions/src/cli.ts
+++ b/packages/graphql-typescript-definitions/src/cli.ts
@@ -4,28 +4,28 @@ import * as yargs from 'yargs';
 import chalk from 'chalk';
 
 import {EnumFormat} from './types';
-import {Builder} from '.';
+import {Builder, SchemaBuild, DocumentBuild} from '.';
 
 const argv = yargs
-  .usage('Usage: $0 <graphql-files> [options]')
-  .option('schema-path', {
-    required: true,
+  .usage('Usage: $0 [options]')
+  .option('watch', {
+    required: false,
+    default: false,
+    type: 'boolean',
+    describe: 'Watch the GraphQL files for changes and re-run the generation',
+  })
+  .option('cwd', {
+    required: false,
+    default: process.cwd(),
     normalize: true,
     type: 'string',
-    describe:
-      'The path to the JSON file containing a schema instrospection query result',
+    describe: 'Working directory to use',
   })
   .option('schema-types-path', {
     required: true,
     normalize: true,
     type: 'string',
     describe: 'The path to output concrete schema types',
-  })
-  .option('watch', {
-    required: false,
-    default: false,
-    type: 'boolean',
-    describe: 'Watch the GraphQL files for changes and re-run the generation',
   })
   .option('add-typename', {
     required: false,
@@ -48,30 +48,50 @@ const argv = yargs
   .help().argv;
 
 const builder = new Builder({
-  graphQLFiles: argv._[0],
-  schemaPath: argv.schemaPath,
+  cwd: argv.cwd,
   schemaTypesPath: argv.schemaTypesPath,
   addTypename: argv.addTypename,
   enumFormat: argv.enumFormat,
 });
 
+const schemas: SchemaBuild[] = [];
+const docs: DocumentBuild[] = [];
+
 const BUILT = chalk.inverse.bold.green(' BUILT ');
 const ERROR = chalk.inverse.bold.red(' ERROR ');
 
-builder.on('start', () => {
+builder.on('start:docs', () => {
+  docs.length = 0;
   console.log();
 });
 
-builder.on('schema:start', () => {
+builder.on('start:schema', () => {
+  schemas.length = 0;
   console.log();
 });
 
-builder.on('schema:end', () => {
-  console.log(`${BUILT} ${argv.schemaTypesPath}`);
+builder.on('build:docs', (doc) => {
+  docs.push(doc);
 });
 
-builder.on('build', ({documentPath, definitionPath}) => {
-  console.log(`${BUILT} ${chalk.dim(documentPath)} → ${definitionPath}`);
+builder.on('build:schema', (schema) => {
+  schemas.push(schema);
+});
+
+builder.on('end:docs', () => {
+  docs
+    .sort(({documentPath: a}, {documentPath: b}) => a.localeCompare(b))
+    .forEach(({documentPath, definitionPath}) => {
+      console.log(`${BUILT} ${chalk.dim(documentPath)} → ${definitionPath}`);
+    });
+});
+
+builder.on('end:schema', () => {
+  schemas
+    .sort(({schemaPath: a}, {schemaPath: b}) => a.localeCompare(b))
+    .forEach(({schemaPath, schemaTypesPath}) => {
+      console.log(`${BUILT} ${chalk.dim(schemaPath)} → ${schemaTypesPath}`);
+    });
 });
 
 builder.on('error', (error) => {
@@ -81,7 +101,7 @@ builder.on('error', (error) => {
   }
   console.log();
 
-  if (!builder.watching) {
+  if (!argv.watch) {
     process.exit(1);
   }
 });
--- a/packages/graphql-typescript-definitions/src/index.ts
+++ b/packages/graphql-typescript-definitions/src/index.ts
@@ -10,11 +10,22 @@ import {
   concatAST,
 } from 'graphql';
 import chalk from 'chalk';
-import {dirname} from 'path';
+import {dirname, isAbsolute, resolve} from 'path';
 import {readJSON, readFile, writeFile, mkdirp} from 'fs-extra';
 import {watch} from 'chokidar';
 import * as glob from 'glob';
-import {compile, Operation, Fragment, AST} from 'graphql-tool-utilities/ast';
+import {
+  getGraphQLConfig,
+  GraphQLProjectConfig,
+  GraphQLConfig,
+} from 'graphql-config';
+import {
+  compile,
+  isOperation,
+  Operation,
+  Fragment,
+  AST,
+} from 'graphql-tool-utilities/ast';
 
 import {printDocument, printSchema} from './print';
 import {EnumFormat} from './types';
@@ -22,80 +33,87 @@ import {EnumFormat} from './types';
 export {EnumFormat};
 
 export interface Options {
-  graphQLFiles: string;
-  schemaPath: string;
-  schemaTypesPath: string;
   addTypename: boolean;
   enumFormat?: EnumFormat;
+  schemaTypesPath: string;
+}
+
+export interface BuilderOptions extends Options {
+  cwd?: string;
 }
 
 export interface RunOptions {
   watch?: boolean;
 }
 
-export interface Build {
+export interface SchemaBuild {
+  schemaPath: string;
+  schemaTypesPath: string;
+}
+
+export interface DocumentBuild {
   documentPath: string;
   definitionPath: string;
   operation?: Operation;
   fragments: Fragment[];
 }
 
-export class Builder extends EventEmitter {
-  watching = false;
-  private globs: string;
-  private schemaPath: string;
-  private schema!: GraphQLSchema;
-  private options: Pick<
-    Options,
-    Exclude<keyof Options, 'graphQLFiles' | 'schemaPath'>
-  >;
-  private documentCache = new Map<string, DocumentNode>();
+type GraphQLDocumentMapByProject = Map<string, Map<string, DocumentNode>>;
 
-  constructor({graphQLFiles, schemaPath, ...options}: Options) {
+export class Builder extends EventEmitter {
+  private options: Options;
+  // workspace graphql configuration
+  // see: https://github.com/prisma/graphql-config
+  private readonly config: GraphQLConfig;
+  private documentMapByProject = new Map<string, Map<string, DocumentNode>>();
+
+  constructor({cwd, ...options}: BuilderOptions) {
     super();
-    this.globs = graphQLFiles;
-    this.schemaPath = schemaPath;
     this.options = options;
+
+    this.config = getGraphQLConfig(cwd ? resolve(cwd) : undefined);
   }
 
   once(event: 'error', handler: (error: Error) => void): this;
-  once(event: 'build', handler: (built: Build) => void): this;
-  once(event: 'start', handler: () => void): this;
-  once(event: 'end', handler: () => void): this;
-  once(event: 'schema:start', handler: () => void): this;
-  once(event: 'schema:end', handler: () => void): this;
+  once(event: 'build:docs', handler: (built: DocumentBuild) => void): this;
+  once(event: 'build:schema', handler: (built: SchemaBuild) => void): this;
+  once(event: 'start:docs', handler: () => void): this;
+  once(event: 'end:docs', handler: () => void): this;
+  once(event: 'start:schema', handler: () => void): this;
+  once(event: 'end:schema', handler: () => void): this;
   once(event: string, handler: (...args: any[]) => void): this {
     return super.once(event, handler);
   }
 
   on(event: 'error', handler: (error: Error) => void): this;
-  on(event: 'build', handler: (built: Build) => void): this;
-  on(event: 'start', handler: () => void): this;
-  on(event: 'end', handler: () => void): this;
-  on(event: 'schema:start', handler: () => void): this;
-  on(event: 'schema:end', handler: () => void): this;
+  on(event: 'build:docs', handler: (built: DocumentBuild) => void): this;
+  on(event: 'build:schema', handler: (built: SchemaBuild) => void): this;
+  on(event: 'start:docs', handler: () => void): this;
+  on(event: 'end:docs', handler: () => void): this;
+  on(event: 'start:schema', handler: () => void): this;
+  on(event: 'end:schema', handler: () => void): this;
   on(event: string, handler: (...args: any[]) => void): this {
     return super.on(event, handler);
   }
 
   emit(event: 'error', error: Error): boolean;
-  emit(event: 'build', built: Build): boolean;
-  emit(event: 'start'): boolean;
-  emit(event: 'end'): boolean;
-  emit(event: 'schema:start'): boolean;
-  emit(event: 'schema:end'): boolean;
+  emit(event: 'build:docs', built: DocumentBuild): boolean;
+  emit(event: 'build:schema', built: SchemaBuild): boolean;
+  emit(event: 'start:docs'): boolean;
+  emit(event: 'end:docs'): boolean;
+  emit(event: 'start:schema'): boolean;
+  emit(event: 'end:schema'): boolean;
   emit(event: string, ...args: any[]): boolean {
     return super.emit(event, ...args);
   }
 
   async run({watch: watchGlobs = false} = {}) {
-    this.watching = watchGlobs;
+    const globs = this.getGlobs();
+    const schemaPaths = this.getSchemaPaths();
 
-    const {globs} = this;
-
-    const update = async (file: string) => {
+    const update = async (filePath: string) => {
       try {
-        await this.updateDocumentForFile(file);
+        await this.updateDocumentForFile(filePath);
       } catch (error) {
         this.emit('error', error);
         return;
@@ -109,18 +127,20 @@ export class Builder extends EventEmitter {
       documentWatcher.on('ready', () => {
         documentWatcher.on('add', update);
         documentWatcher.on('change', update);
-        documentWatcher.on('unlink', async (file: string) => {
-          this.removeDocumentForFile(file);
+        documentWatcher.on('unlink', async (filePath: string) => {
+          this.removeDocumentForFile(filePath);
           await this.generateDocumentTypes();
         });
       });
 
-      const schemaWatcher = watch(this.schemaPath);
+      const schemaWatcher = watch(schemaPaths);
       schemaWatcher.on('ready', () => {
-        schemaWatcher.on('change', async () => {
+        schemaWatcher.on('change', async (schemaPath: string) => {
           try {
-            await this.updateSchema();
-            await this.generateSchemaTypes();
+            this.emit('start:schema');
+            await this.updateSchemaAndGenerateTypes(schemaPath);
+            this.emit('end:schema');
+
             await this.generateDocumentTypes();
           } catch (error) {
             // intentional noop
@@ -130,8 +150,11 @@ export class Builder extends EventEmitter {
     }
 
     try {
-      await this.updateSchema();
-      await this.generateSchemaTypes();
+      this.emit('start:schema');
+      await Promise.all(
+        schemaPaths.map(this.updateSchemaAndGenerateTypes.bind(this)),
+      );
+      this.emit('end:schema');
     } catch (error) {
       this.emit('error', error);
       return;
@@ -139,7 +162,10 @@ export class Builder extends EventEmitter {
 
     try {
       await Promise.all(
-        glob.sync(globs).map(this.updateDocumentForFile.bind(this)),
+        globs
+          .map((pattern) => glob.sync(pattern))
+          .reduce((patterns, globbed) => patterns.concat(globbed), [])
+          .map(this.updateDocumentForFile.bind(this)),
       );
     } catch (error) {
       this.emit('error', error);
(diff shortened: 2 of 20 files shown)


patsissons:
no tests yet, any ideas on how best to test this let me know. otherwise i'll probably figure something out tomorrow.

patsissons:
this will probably collide pretty harshly with #13

packages/graphql-typescript-definitions/src/cli.ts
```diff
@@ -47,8 +48,12 @@ const argv = yargs
   })
   .help().argv;
 
+const cwd = process.cwd();
+const graphQLFiles = relative(cwd, argv._[0]);
```
patsissons: we do this so that we can get protect against absolute glob paths.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -10,21 +10,30 @@ import {
...
 import {compile, Operation, Fragment, AST} from 'graphql-tool-utilities/ast';
 
 import {printDocument, printSchema} from './print';
 import {EnumFormat} from './types';
 
 export {EnumFormat};
 
-export interface Options {
+export interface CLIOptions {
+  cwd: string;
```
patsissons: this `cwd` member is a bit of a _flag_ option, but also could be useful if we wanted to add custom cwd support to the cli. basically i needed a way to short circuit the existing SK usage so that it defaulted to using the workspace `.graphqlconfig` instead of interpreting the usage as a CLI invocation.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -40,22 +49,48 @@ export interface Build {
   fragments: Fragment[];
 }
 
+// projectName => { filePath => document }
+export type GraphQLProjectDocuments = Map<string, Map<string, DocumentNode>>;
```
patsissons: this type represents our internal processing structure, a nested map.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -222,16 +293,17 @@ export class Builder extends EventEmitter {
...
   }
 
-  private async updateSchema() {
+  private async updateSchema(schemaPath: string) {
     try {
-      const schemaJSON = await readJSON(this.schemaPath);
-      this.schema = buildClientSchema(schemaJSON.data);
+      const schemaJSON = await readJSON(schemaPath);
+      const schema = buildClientSchema(schemaJSON.data);
+      await this.generateSchemaTypes(schemaPath, schema);
```
patsissons: i noticed we could generate types in the `updateSchema` call because we were doing that after the call anyways.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -240,21 +312,108 @@ export class Builder extends EventEmitter {
...
   }
 
-  private async updateDocumentForFile(file: string) {
-    const contents = await readFile(file, 'utf8');
-    if (contents.trim().length === 0) {
-      return;
+  private async updateDocumentForFile(filePath: string) {
+    const project = this.config.getConfigForFile(filePath);
+
+    if (project) {
```
patsissons: i think this should always be true, so we may want to `error` on the else? technically our file paths should be coming from `includes` globs so they should always belong to a project. This was added because of the absolute path globbing issue.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -240,21 +312,108 @@ export class Builder extends EventEmitter {
...
+  private getProjectDocuments(): GraphQLProjectDocuments {
+    return Array.from(this.documentCache.entries())
+      .map(([filePath, document]) => ({
+        document,
+        filePath,
+        project: this.config.getConfigForFile(filePath),
+      }))
+      .reduce((groups, {document, filePath, project}) => {
+        if (project) {
+          const group = groups.get(project.projectName || '');
```
patsissons: `projectName` can be null if there are no projects defined in the config (i.e. single unwrapped project). empty string seemed like the safe option to default to here. We should always have a `projectName` set in practice (for our use cases). I did notice some awkwardness with `graphql-config` when using the single unwrapped project format, so I recommend against it.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -240,21 +312,108 @@ export class Builder extends EventEmitter {
...
+  );
+}
+
+function getSchemaTypesPath(project: GraphQLProjectConfig) {
+  if (typeof project.extensions.schemaTypesPath === 'string') {
+    return project.extensions.schemaTypesPath;
+  }
+
+  if (project.schemaPath) {
+    return project.schemaPath.replace(/\.json$/, '.types.ts');
```
patsissons: this was just a sane default choice for how to format schema types, open to suggestions if we want to change it.

tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "noUnusedParameters": true,
     "experimentalDecorators": true,
     "target": "es2015",
-    "lib": ["es2015", "es2016", "es2017", "esnext.asynciterable"]
+    "lib": ["dom", "es2015", "es2016", "es2017", "esnext.asynciterable"]
```
patsissons: `dom` lib is required by `graphql-config` for its internal mechanism for fetching remote schemas.

packages/graphql-typescript-definitions/src/index.ts
TheMallen: maybe we should have a validation step that errors out when when either `project` or `projectName` is falsy?

patsissons: i thought about that but i was hesitant because the `.graphqlconfig` scheme technically supports it (and they will likely fix the bugs that currently exist sometime soon)

packages/graphql-typescript-definitions/src/cli.ts
TheMallen: This might actually want a comment in code since it's not very obvious

packages/graphql-typescript-definitions/src/index.ts
patsissons: I just noticed that this fails if you provide a `*.graphql` schema path

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -40,22 +49,48 @@ export interface Build {
   fragments: Fragment[];
 }
 
+// projectName => { filePath => document }
```
TheMallen: 🔥

packages/graphql-typescript-definitions/src/index.ts
patsissons: turns out you probably need this if you want the cli to use a config from a specific path

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -40,22 +49,48 @@ export interface Build {
...
+                schemaTypesPath,
+              },
+            },
+          },
+        },
+        // graphql-config depends on the location of this file so it needs to be sane
+        resolve(cwd, '.graphqlconfig'),
+      );
+    } else {
+      this.config = getGraphQLConfig();
```
patsissons: I should probably be checking `graphQLFiles` here to see if it is a `*.graphqlconfig` path and then construct the config accordingly.

packages/graphql-typescript-definitions/src/index.ts
TheMallen: I think that it is better to leave `generateSchemaTypes` out of this function and add it back to where they are both invoked, just to keep levels of abstraction more consistent across the individual task functions like this, and to make the main flows a bit more grokkable.

packages/graphql-typescript-definitions/src/index.ts
patsissons: yea probably don't need that any longer, it was helpful during the build phase 😁

packages/graphql-typescript-definitions/src/index.ts
patsissons: one of the motivations was that it made the `Promise.all(...)` a lot cleaner. but i guess i could just add a new `updateSchemaAndGenerateTypes` wrapper function that calls both `updateSchema` and `generateSchemaTypes`.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -149,36 +193,60 @@ export class Builder extends EventEmitter {
     await this.generateDocumentTypes();
   }
 
-  private async generateSchemaTypes() {
+  private async generateSchemaTypes(schemaPath: string, schema: GraphQLSchema) {
```
patsissons: this function needs to emit the `schemaPath` in `schema:end`

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -130,16 +170,20 @@ export class Builder extends EventEmitter {
...
     } catch (error) {
       this.emit('error', error);
       return;
     }
 
     try {
       await Promise.all(
-        glob.sync(globs).map(this.updateDocumentForFile.bind(this)),
+        globs
+          .map((pattern) => glob.sync(pattern))
```
patsissons: this globbing assumes the workspace and process cwd are the same, we can override the glob cwd with options.

patsissons:
NOTE: there is a bit of a determinism issue with CI because of the way `Promise.all` works. We can (and do consistently) receive completed events in a different order than they were dispatched (notably, queries processing faster than fragments). We shouldn't see this flip-flopping much in CI but it should be noted. If it is possible to reduce concurrency when testing that would resolve non-determinism at the cost of some performance.

packages/graphql-typescript-definitions/src/cli.ts
```diff
@@ -2,21 +2,28 @@
...
-    required: true,
+    required: false,
     normalize: true,
     type: 'string',
     describe:
       'The path to the JSON file containing a schema instrospection query result',
   })
   .option('schema-types-path', {
-    required: true,
+    required: false,
```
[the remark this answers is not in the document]
patsissons: the cli supports one or multiple schemas through the `.graphqlconfig` or alternatively supports a single schema through these cli params (and no `.graphqlconfig`). The happy path is to always use the config, so that's why these were no longer required. Basically this is just for backwards compat.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -33,37 +42,73 @@ export interface RunOptions {
...
+}
+
 export interface Build {
   documentPath: string;
   definitionPath: string;
   operation?: Operation;
   fragments: Fragment[];
 }
 
+export type GraphQLProjectDocuments = Map<string, Map<string, DocumentNode>>;
```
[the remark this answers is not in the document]
patsissons: probably not, i was going to add more tests and thought i may need it exported for those tests.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -33,37 +42,73 @@ export interface RunOptions {
...
-  >;
+  private options: Options;
+  // workspace graphql configuration
+  // see: https://github.com/prisma/graphql-config
+  private readonly config: GraphQLConfig;
   private documentCache = new Map<string, DocumentNode>();
 
-  constructor({graphQLFiles, schemaPath, ...options}: Options) {
+  constructor({
+    cwd,
```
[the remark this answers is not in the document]
patsissons: being undefined is actually a valid value, it means that we compute cwd based on the location of the `.graphqlconfig` path

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -240,21 +344,118 @@ export class Builder extends EventEmitter {
...
+            group.set(filePath, document);
+          } else {
+            groups.set(
+              project.projectName || '',
+              new Map<string, DocumentNode>([[filePath, document]]),
+            );
+          }
+        }
+        return groups;
+      }, new Map<string, Map<string, DocumentNode>>());
```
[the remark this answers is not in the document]
patsissons: yup! `graphql-config` uses `projectName` instead of `schemaName` for terminology, but effectively the same.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -240,21 +344,118 @@ export class Builder extends EventEmitter {
...
+  if (project) {
+    return project;
+  }
+
+  throw new Error(
+    `No project defined in .graphqlconfig for schema ${schemaPath}`,
+  );
+}
+
+function getSchemaTypesPath(project: GraphQLProjectConfig) {
```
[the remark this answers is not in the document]
patsissons: I think i can make that work

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -240,21 +344,118 @@ export class Builder extends EventEmitter {
...
-  private async updateDocumentForFile(file: string) {
-    const contents = await readFile(file, 'utf8');
-    if (contents.trim().length === 0) {
-      return;
+  private async updateDocumentForFile(filePath: string) {
+    const project = this.config.getConfigForFile(filePath);
+
+    if (project) {
+      const contents = await readFile(
+        getFullFilePath(this.config, filePath),
```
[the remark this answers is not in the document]
patsissons: this is basically to deal with that globbing issue. I was observing pathing issues when i ran the tool for a cwd that was not the process' cwd. so this `getFullFilePath` ensures that we always operate on a proper path to the file.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -149,36 +204,61 @@ export class Builder extends EventEmitter {
...
-          )} is declared in:\n ${files.join('\n ')}`,
+          this.emit('error', new Error(message));
+        });
+      }
+    });
+
+    await Promise.all(
+      Array.from(projectDocuments.entries()).map(([projectName, documents]) => {
+        return this.generateProjectDocumentTypes(
+          this.config.getProjectConfig(projectName),
```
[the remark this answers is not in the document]
patsissons: I'll see if i can make this alteration

packages/graphql-typescript-definitions/src/cli.ts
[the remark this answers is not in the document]
patsissons: based on your other comment, i'll make this required again.

packages/graphql-typescript-definitions/src/cli.ts
```diff
@@ -2,21 +2,28 @@
...
 
 const argv = yargs
   .usage('Usage: $0 <graphql-files> [options]')
+  .option('cwd', {
+    required: false,
+    normalize: true,
+    type: 'string',
+    describe: 'Working directory to use',
+  })
   .option('schema-path', {
```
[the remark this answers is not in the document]
patsissons: this option is still used to process a schema without a `.graphqlconfig` file

tsconfig.json
lemonmade: I think you probably want the node lib instead in that case, which I think includes fetch typings? Either that or one of the polyfill libraries' typings should be defining those?

patsissons: there is some weirdness around this one. the node libs are also using the dom lib for these definitions. i need to dig a bit further to make sure my world view is sane.

patsissons: The world view is a bit troubling, [`graphql-config`](https://github.com/prisma/graphql-config/blob/master/tsconfig.json#L8) and [`graphql-request`](https://github.com/prisma/graphql-request/blob/master/tsconfig.json#L11) (dependency of `graphql-config`) are both loading the dom lib. I agree that we should not be doing this (and [so does mhegazy](https://github.com/Microsoft/TypeScript/issues/4948#issuecomment-281513657)). I added the dom lib originally because [someone suggested it](https://github.com/prisma/graphql-request/issues/26#issuecomment-354482903) without really thinking about the consequences. I think we may need to inject ambient types to cover the missing two until `graphql-request` [removes its dependence](https://github.com/prisma/graphql-request/pull/100) on the dom lib. It looks like internally it is using [`cross-fetch/polyfill`](https://github.com/lquixada/cross-fetch) which is using [`node-fetch`](https://github.com/lquixada/cross-fetch/blob/master/dist/node-ponyfill.js#L1) internally. So i'll put some effort into producing some ambient types based on the [`@types/node-fetch`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node-fetch/index.d.ts) exports.

patsissons:
@lemonmade have a look at the [ambient type injection module](https://github.com/Shopify/graphql-tools-web/blob/b165138506daf34be2d31938b6980610de9050f1/packages/graphql-typescript-definitions/src/ambient-graphql-request-patch.ts). it allows us to circumvent the dom lib requirement from `graphql-config`.

packages/graphql-typescript-definitions/src/cli.ts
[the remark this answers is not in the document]
lemonmade: My proposal would be to treat this as a breaking change and require any project using this to use the .graphqlconfig file. Then we simplify the code and the documentation needed to successfully use this tool/ Sewing Kit for GraphQL purposes.

packages/graphql-typescript-definitions/src/cli.ts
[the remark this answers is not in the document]
lemonmade: Is this going to be required? I feel this is disconnected from the graphql config — that is for "where are my schema and documents", this is for "where do you want your types".

packages/graphql-typescript-definitions/src/cli.ts
```diff
@@ -47,31 +54,62 @@ const argv = yargs
   })
   .help().argv;
 
+if (typeof argv.cwd === 'string') {
+  process.chdir(argv.cwd);
+}
```
lemonmade: I think this at least needs some comments to explain, IMO it is generally not good practice to do this

packages/graphql-typescript-definitions/src/cli.ts
```diff
@@ -47,31 +54,62 @@ const argv = yargs
...
+
+// convert the graphql globs to a relative path (if provided)
+// this is to deal with an issue with graphql-config and absolute path globbing
+// see: https://github.com/stephen/vscode-graphql/issues/23
+const graphQLFiles = argv._[0] && relative(process.cwd(), argv._[0]);
```
lemonmade: Here is another one that we could conceivably just do the breaking change and only allow it to be defined through the GraphQL config

packages/graphql-typescript-definitions/src/cli.ts
```diff
@@ -47,31 +54,62 @@ const argv = yargs
...
 
-builder.on('schema:start', () => {
+builder.on('start:schema', () => {
+  schemas.length = 0;
   console.log();
 });
 
-builder.on('schema:end', () => {
-  console.log(`${BUILT} ${argv.schemaTypesPath}`);
+builder.on('build', (doc) => {
```
lemonmade: Can you open a follow-up issue to rename some of these events, they don't really make sense now that we've added the schema ones

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -10,88 +10,128 @@ import {
...
+    graphQLFiles,
+    schemaPath,
+    ...options
+  }: Options & Partial<CLIOptions>) {
     super();
-    this.globs = graphQLFiles;
-    this.schemaPath = schemaPath;
     this.options = options;
+
+    if (cli && graphQLFiles && schemaPath) {
```
lemonmade: Why does it matter if the cli is doing this? IMO this is very smelly

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -130,16 +178,23 @@ export class Builder extends EventEmitter {
...
       return;
     }
 
     try {
       await Promise.all(
-        glob.sync(globs).map(this.updateDocumentForFile.bind(this)),
+        globs
+          .map((pattern) => glob.sync(pattern, {cwd: this.config.configDir}))
+          .reduce((patterns, globbed) => patterns.concat(globbed), [])
+          .sort()
```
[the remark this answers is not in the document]
lemonmade: Reping

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -240,19 +338,119 @@ export class Builder extends EventEmitter {
...
+    if (project) {
+      return [project];
+    }
+
+    this.emit('error', new Error(`No projects configured in .graphqlconfig`));
+
+    return [];
   }
+
+  private getGlobs() {
```
lemonmade: I still think these make more sense as real getters but it's not a huge deal

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -240,19 +338,119 @@ export class Builder extends EventEmitter {
...
+      return Object.values(projects);
+    }
+
+    const project = this.config.getProjectConfig();
+
+    if (project) {
+      return [project];
+    }
+
+    this.emit('error', new Error(`No projects configured in .graphqlconfig`));
```
lemonmade: Personally, I'd rather validate this in the constructor and just fail as early as possible

packages/graphql-typescript-definitions/src/cli.ts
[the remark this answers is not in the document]
patsissons: ah ok, i thought we would want to maintain backwards compat. if we can change this without that then I have no problem getting rid of the extra code designed for cli usage.

packages/graphql-typescript-definitions/src/cli.ts
patsissons: if you're not on board with it, i can go back to processing paths based on this value (and default to `process.cwd()`).

packages/graphql-typescript-definitions/src/cli.ts
patsissons: yea i wanted to maintain compat, but really i'll just fix them all if we are not concerned about compat.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -10,92 +10,137 @@ import {
...
+} from 'graphql-tool-utilities/ast';
+import {
+  Headers as FetchHeaders,
+  RequestInit as FetchRequestInit,
+} from 'node-fetch';
 
 import {printDocument, printSchema} from './print';
 import {EnumFormat} from './types';
 
+// adding some type definitions to the global namespace so that graphql-request
```
lemonmade: I just want to make sure — this doesn't end up conflicting with other TS projects consuming it (notably, Sewing Kit)?

patsissons: that's a good point. AFAIK, it should not because only the `dom` lib places these types in the global namespace. anything node related should be exported via the `node-fetch` module. The truth is that it _could_ conflict with anything that expects these types coming from the `dom` lib, but perhaps this *is* the appropriate response as we probably want to fix those usages anyways. That said, i'll 🎩 against SK right now and confirm.

PR #21, opened 2018-09-20
updating graphql-validate-fixtures to support graphql-config

patsissons:
Updates this tool to use `graphql-config` for resolving schemas. Fixtures are mapped to operation names by the name of the directory they live within (or by using the `@operation` pragma if the directory name is not suitable). The workspace configuration is loaded and all operations all included files per schema are compiled into AST, which allows us to connect between fixture and operation, and by extension fixture and schema.

The `schemaOnly` (`--schema-only`) flag has been removed as it is no longer purposeful since we need to compile operations into AST for mapping purposes regardless of whether the flag is set (and the flag's main purpose was to speed boost by not performing a compilation). This means a lot of related functions were also removed because they were no longe being referenced.

It is possible for two operations to share the same name across multiple schemas, which can result in a fixture having an ambiguous mapping to a schema. In a future version, a new `@project` pragma will be introduced to directly map a fixture to a schema. Currently if this scenario occurs an _Ambiguous operation name_ error is produced as a validation error.

### DIFF
--- a/packages/graphql-typescript-definitions/src/index.ts
+++ b/packages/graphql-typescript-definitions/src/index.ts
@@ -20,18 +20,18 @@ import {
   GraphQLConfig,
 } from 'graphql-config';
 import {
+  getGraphQLFilePath,
+  getGraphQLProjectForSchemaPath,
+  getGraphQLProjects,
+  getGraphQLSchemaPaths,
+} from 'graphql-tool-utilities';
+import {
   compile,
   isOperation,
   Operation,
   Fragment,
   AST,
 } from 'graphql-tool-utilities/ast';
-import {
-  getGraphQLFilePath,
-  getGraphQLProjectForSchemaPath,
-  getGraphQLProjects,
-  getGraphQLSchemaPaths,
-} from 'graphql-tool-utilities/config';
 
 import {printDocument, printSchema} from './print';
 import {EnumFormat} from './types';
--- a/packages/graphql-validate-fixtures/README.md
+++ b/packages/graphql-validate-fixtures/README.md
@@ -23,13 +23,26 @@ In order to associate a fixture with a GraphQL query or mutation in your app, yo
 
 Once this is done, you can validate your fixtures using the CLI or Node.js API.
 
+### Operation
+
+On startup this tool performs the following actions:
+
+* Loads all schemas
+* Discovers all operations belonging to each schema
+* Discovers all fixtures and infers operation names as described [above](#Usage)
+* Validates fixtures against the operation with a matching name
+  * Reports operation not found error if no schema matches
+  * Reports ambiguous operation name error if more than one schema matches
+
+### Configuration
+
+This tool reads schema information from a [`.graphqlconfig`](https://github.com/Shopify/graphql-tools-web/tree/master/packages/graphql-tool-utilities#configuration) file in the project root.
+
 ### CLI
 
 ```sh
-# Must provide a list of fixtures as the first argument, as well
-# as flags for the path to the schema (`--schema-path`) and operations
-# (`--operation-paths`, can be a glob pattern)
-yarn run graphql-validate-fixtures 'src/**/fixtures/**/*.graphql.json' --schema-path 'build/schema.json' --operation-paths 'src/**/*.graphql'
+# Must provide a list of fixtures as the first argument
+yarn run graphql-validate-fixtures 'src/**/fixtures/**/*.graphql.json'
 ```
 
 ### Node
@@ -38,12 +51,9 @@ yarn run graphql-validate-fixtures 'src/**/fixtures/**/*.graphql.json' --schema-
 const {evaluateFixtures} = require('graphql-validate-fixtures');
 evaluateFixtures({
   fixturePaths: ['test/fixtures/one.json', 'test/fixtures/two.json'],
-  schemaPath: 'build/schema.json',
-  operationPaths: ['src/Home/Home.graphql'],
-})
-  .then((results) => {
-    // See the TypeScript definition file for more details on the
-    // structure of the `results`
-    results.forEach((result) => console.log(result));
-  });
+}).then((results) => {
+  // See the TypeScript definition file for more details on the
+  // structure of the `results`
+  results.forEach((result) => console.log(result));
+});
 ```
--- a/packages/graphql-validate-fixtures/src/cli.ts
+++ b/packages/graphql-validate-fixtures/src/cli.ts
@@ -9,25 +9,12 @@ import {evaluateFixtures} from '.';
 
 const argv = yargs
   .usage('Usage: $0 <fixtures> [options]')
-  .option('schema-path', {
-    required: true,
-    normalize: true,
-    type: 'string',
-    describe:
-      'The path to the JSON file containing a schema instrospection query result',
-  })
-  .option('operation-paths', {
+  .option('cwd', {
     required: false,
+    default: process.cwd(),
     normalize: true,
     type: 'string',
-    describe:
-      'The glob pattern for GraphQL queries and fragments to compare against',
-  })
-  .option('schema-only', {
-    type: 'boolean',
-    default: false,
-    describe:
-      'Validate fixtures only against the GraphQL schema (and not any query or mutation documents)',
+    describe: 'Working directory where the .graphqlconfig is located',
   })
   .option('show-passes', {
     type: 'boolean',
@@ -37,18 +24,9 @@ const argv = yargs
   })
   .help().argv;
 
-const hasOperationPaths = Boolean(argv.operationPaths);
-
-evaluateFixtures(
-  {
-    fixturePaths: glob.sync(argv._[0]),
-    operationPaths: hasOperationPaths ? glob.sync(argv.operationPaths) : [],
-    schemaPath: argv.schemaPath,
-  },
-  {
-    schemaOnly: argv.schemaOnly || !hasOperationPaths,
-  },
-)
+evaluateFixtures(glob.sync(argv._[0]), {
+  cwd: argv.cwd,
+})
   .then((evaluations) => {
     let passed = 0;
     let failed = 0;
@@ -57,7 +35,7 @@ evaluateFixtures(
     console.log();
 
     evaluations.forEach((evaluation) => {
-      const relativePath = relative(process.cwd(), evaluation.fixturePath);
+      const relativePath = relative(argv.cwd, evaluation.fixturePath);
       const formattedPath = `${chalk.dim(
         relativePath.replace(basename(relativePath), ''),
       )}${chalk.bold(basename(relativePath))}`;
--- a/packages/graphql-validate-fixtures/src/index.ts
+++ b/packages/graphql-validate-fixtures/src/index.ts
@@ -1,68 +1,64 @@
 import {readFile, readJSON} from 'fs-extra';
-import {isAbsolute, resolve} from 'path';
-import {
-  GraphQLSchema,
-  buildClientSchema,
-  Source,
-  parse,
-  concatAST,
-} from 'graphql';
+import {resolve} from 'path';
+import glob from 'glob';
+import {Source, parse, concatAST, GraphQLSchema} from 'graphql';
+import {getGraphQLConfig, GraphQLProjectConfig} from 'graphql-config';
+import {getGraphQLFilePath, getGraphQLProjects} from 'graphql-tool-utilities';
 import {compile} from 'graphql-tool-utilities/ast';
+import {promisify} from 'util';
 
 import {
-  validateFixtureAgainstAST,
-  validateFixtureAgainstSchema,
-  Validation,
   Fixture,
+  Validation,
+  validateFixture,
+  getOperationForFixture,
+  MissingOperationError,
+  AmbiguousOperationNameError,
+  GraphQLProjectAST,
 } from './validate';
 
-export interface Paths {
-  fixturePaths: string[];
-  operationPaths?: string[];
-  schemaPath: string;
-}
-
 export interface Options {
-  schemaOnly?: boolean;
+  cwd: string;
 }
 
 export interface Evaluation extends Validation {
-  fixturePath: string;
   scriptError?: Error;
 }
 
 export async function evaluateFixtures(
-  {fixturePaths, operationPaths = [], schemaPath}: Paths,
-  {schemaOnly = false}: Options = {},
+  fixturePaths: string[],
+  {cwd}: Options,
 ): Promise<Evaluation[]> {
-  let schema: GraphQLSchema;
+  const config = getGraphQLConfig(resolve(cwd));
 
-  try {
-    const schemaJSON = await readJSON(schemaPath, {encoding: 'utf8'});
-    schema = buildClientSchema(schemaJSON.data);
-  } catch (error) {
-    throw new Error(
-      `Error parsing '${schemaPath}':\n\n${error.message.replace(
-        /Syntax Error GraphQL \(.*?\) /,
-        '',
-      )}`,
-    );
-  }
-
-  if (schemaOnly) {
-    return runForEachFixture(fixturePaths, (fixture) =>
-      validateFixtureAgainstSchema(fixture, schema),
-    );
-  }
-
-  const sources = await Promise.all(
-    operationPaths.map(
-      async (operationPath) =>
-        new Source(await readFile(operationPath, 'utf8'), operationPath),
-    ),
+  const projectASTCollection = await Promise.all(
+    getGraphQLProjects(config).map(getOperationsForProject),
   );
+
+  return runForEachFixture(fixturePaths, (fixture) =>
+    evaluateFixture(fixture, projectASTCollection),
+  );
+}
+
+async function getOperationsForProject(
+  projectConfig: GraphQLProjectConfig,
+): Promise<GraphQLProjectAST> {
+  const operationPaths = (await Promise.all(
+    projectConfig.includes.map((include) =>
+      promisify(glob)(getGraphQLFilePath(projectConfig, include), {
+        ignore: projectConfig.excludes.map((exclude) =>
+          getGraphQLFilePath(projectConfig, exclude),
+        ),
+      }),
+    ),
+  )).flatMap((filePaths) => filePaths);
+
+  const operationSources = await Promise.all(
+    operationPaths.map(loadOperationSource),
+  );
+
   const document = concatAST(
-    sources.map((source) => {
+    operationSources.map((source) => {
       try {
         return parse(source);
       } catch (error) {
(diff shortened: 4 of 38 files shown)


packages/graphql-validate-fixtures/src/index.ts
```diff
@@ -1,68 +1,148 @@
...
+    const [{operationName}] = validFixtureOperations;
+    const projectNames = projectOperations
+      .map(({config}) => config.resolveProjectName())
+      .join(', ');
+
+    return {
+      fixturePath: fixture.path,
+      validationErrors: [],
+      scriptError: new Error(
+        `Ambiguous operation name '${operationName}' in '${
```
GoodForOneFare: Is there a test for this behaviour?

packages/graphql-validate-fixtures/test/fixtures/all-clear/.graphqlconfig.yml
```diff
@@ -0,0 +1,4 @@
+schemaPath: schema.json
```
GoodForOneFare: I'd like a similar test for a multi-project config.

packages/graphql-validate-fixtures/README.md
```diff
@@ -23,13 +23,65 @@ In order to associate a fixture with a GraphQL query or mutation in your app, yo
 
 Once this is done, you can validate your fixtures using the CLI or Node.js API.
 
+### Configuration
+
+This tool consumes a [`.graphqlconfig`](https://github.com/prisma/graphql-config) file, typically placed in the root of the project. This configuration file is compatible with the [VSCode extension](https://github.com/prisma/vscode-graphql) to provide syntax highlighting and autocomplete suggestions. The configuration file supports a single nameless project or multiple named projects, with each project linked to a schema and a set of include and exclude globbing patterns. Upon processing a fixture file, the fixutres opretaion name will be inferred (as described above) and that operation name will be used as a lookup in the available schemas that have been loaded. If a single schema is matched, then the fixture is validated against the compiled AST of the operations within that schema (who themselves are validated against the schema). If a signle schema is not matched, then either an operation not found error or a ambiguous operation name error is thrown.
```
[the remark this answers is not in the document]
patsissons: reduced to G10, but probably not going to reduce much further due to the complicated nature of this tool.

packages/graphql-validate-fixtures/src/cli.ts
```diff
@@ -9,25 +9,12 @@ import {evaluateFixtures} from '.';
...
     type: 'string',
-    describe:
-      'The glob pattern for GraphQL queries and fragments to compare against',
-  })
-  .option('schema-only', {
-    type: 'boolean',
-    default: false,
-    describe:
-      'Validate fixtures only against the GraphQL schema (and not any query or mutation documents)',
+    describe: 'Working directory to use',
```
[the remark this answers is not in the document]
patsissons: going with `Working directory where the .graphqlconfig is located`

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -340,38 +346,28 @@ export class Builder extends EventEmitter {
...
-
-    return [this.config.getProjectConfig()];
+    return getGraphQLProjects(this.config);
   }
 
   private getGlobs() {
     return this.getProjects().reduce<string[]>((globs, project) => {
       return globs.concat(
-        project.includes.map(getFilePath.bind(this, this.config.configDir)),
+        project.includes.map(getGraphQLFilePath.bind(this, this.config)),
```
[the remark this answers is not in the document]
patsissons: haha mistakenly bound to `this` when i wrote it the first time.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -340,38 +346,28 @@ export class Builder extends EventEmitter {
...
   }
 
   private removeDocumentForFile(filePath: string) {
     const project = this.config.getConfigForFile(filePath);
 
     if (project) {
-      const projectName = project.projectName || '';
-      const documents = this.documentMapByProject.get(projectName);
+      const documents = this.documentMapByProject.get(
+        project.resolveProjectName(''),
```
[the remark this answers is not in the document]
patsissons: This is a bit of a strange one. Normally if you don't provide a default it will use our _sane_ default `projectName` of `GraphQL`. For this tool I didn't want to use a _sane_ default because I care if `projectName` is `falsy` when we are producing the schema types file name.

packages/graphql-tool-utilities/augmentations.ts
```diff
@@ -0,0 +1,16 @@
...
+
+declare module 'graphql-config/lib/GraphQLProjectConfig' {
+  interface GraphQLProjectConfig {
+    resolveProjectName(defaultName?: string): string;
+  }
+}
+
+function resolveProjectName(
+  this: GraphQLProjectConfig,
+  defaultName = 'GraphQL',
```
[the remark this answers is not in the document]
patsissons: `GraphQL` was chris' suggestion, i'm not particularly married to anything (i used to have it as `unnamed` haha). I'll have a closer look at the code and make sure that we are emitting the `schema.graphql` properly (it should share the same name as the `schema.json`).

packages/graphql-validate-fixtures/src/cli.ts
```diff
@@ -9,25 +9,12 @@ import {evaluateFixtures} from '.';
...
   .usage('Usage: $0 <fixtures> [options]')
-  .option('schema-path', {
-    required: true,
-    normalize: true,
-    type: 'string',
-    describe:
-      'The path to the JSON file containing a schema instrospection query result',
-  })
-  .option('operation-paths', {
+  .option('cwd', {
```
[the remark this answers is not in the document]
patsissons: this was mostly just so that all the tooling was consistent with `cwd` being used as a param.

packages/graphql-validate-fixtures/src/index.ts
```diff
@@ -1,68 +1,148 @@
...
+interface GraphQLProjectOperations {
+  config: GraphQLProjectConfig;
+  ast: AST;
+}
+
+async function getOperationsForProject(
+  projectConfig: GraphQLProjectConfig,
+): Promise<GraphQLProjectOperations> {
+  const operationPaths = projectConfig.includes
+    .map<string>(getGraphQLFilePath.bind(undefined, projectConfig))
```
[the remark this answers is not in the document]
patsissons: is that preferred? both expressions should be equivalent, so i can go either way.

packages/graphql-validate-fixtures/src/index.ts
patsissons: I'll write some additional tests today to cover these new scenarios.

patsissons: should these be script errors or validation errors?

packages/graphql-validate-fixtures/src/index.ts
```diff
@@ -1,68 +1,148 @@
...
-      validateFixtureAgainstSchema(fixture, schema),
-    );
+  if (validFixtureOperations.length === 0) {
+    let errors = fixtureOperations
+      .filter(({operation}) => operation != null)
+      .map(({error}) => error);
+
+    if (errors.length === 0) {
+      // we couldn't find an error in a valid fixture, fallback on all errors
+      errors = fixtureOperations.map(({error}) => error);
```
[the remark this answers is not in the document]
patsissons: i made the change, i think it looks much better.

```
[Error: Could not find a matching operation (looked for 'Missing' based on the fixture’s directory name). Make sure to put your fixture in a folder named the same as the operation, or add an '@operation' key indicating the operation.. Available operations: MyA, MyB],
```

packages/graphql-validate-fixtures/src/index.ts
```diff
@@ -1,68 +1,168 @@
...
 }
 
 export async function evaluateFixtures(
-  {fixturePaths, operationPaths = [], schemaPath}: Paths,
-  {schemaOnly = false}: Options = {},
+  fixturePaths: string[],
+  {cwd}: Options,
 ): Promise<Evaluation[]> {
-  let schema: GraphQLSchema;
+  const config = getGraphQLConfig(cwd ? resolve(cwd) : undefined);
```
cartogram: Doesn't look like this ternary is needed, cwd will always be truthy.

patsissons: good catch! It was previously not the case, but now i can remove this check.

packages/graphql-validate-fixtures/test/fixtures/ambiguous-operation/fixtures/projectA/another-another-fixture.json
```diff
@@ -0,0 +1,4 @@
+{
```
GoodForOneFare: I know the boilerplate is awful for these tests, but there's too much happening in this fixture.  I have to read the snapshot file to be sure that `another-fixture` and `another-another-fixture` independently cause errors.

packages/graphql-validate-fixtures/src/index.ts
```diff
@@ -1,68 +1,168 @@
...
+    const [{operationName}] = validFixtureOperations;
+    const projectNames = projectOperations
+      .map(({config}) => config.resolveProjectName())
+      .join(', ');
+
+    return {
+      fixturePath: fixture.path,
+      validationErrors: [
+        new Error(
+          `Ambiguous operation name '${operationName}' in '${
```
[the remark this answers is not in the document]
patsissons: We could run through all schemas and check for dupes, but it's probably not worthwhile since we wouldn't really save much time and this should be a super rare occurrence. Right now there isn't much you can do if this happens other than change the name. Chris and I discussed the option of using a `@schema` pragma similar to the `@operation` pragma to override ambiguities, this would solve the problem pretty easily if it ever did happen. In reality, this should never happen because myself and the guidance I will provide for any other teams looking to put their schema in web will be to prefix your operations (i.e., 
`CoreCapitalIndexQuery` and `BourgeoisCapitalIndexQuery`).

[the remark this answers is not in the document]
GoodForOneFare: I was more concerned about readability than speed.  If this is super rare, and can be shuffled into a helper function that nobody ever looks at, all the better.

That said, not a dealbreaker.

packages/graphql-validate-fixtures/test/fixtures/ambiguous-operation/fixtures/projectA/another-another-fixture.json
patsissons: these multi-project fixtures are just complex by the nature of what they are testing. i can simplify them down a little so that they only test one type of ambiguous detection, but that would multiple the added files even further. For this particular fixture, the min required content is `graphqconfig`, two schemas, two queries, and two fixtures. The extra files are just testing different methods of detecting the ambiguity. That is perhaps not necessary given we have other tests that cover how fixtures map to operation names.

packages/graphql-validate-fixtures/src/index.ts
```diff
@@ -1,68 +1,168 @@
...
+      .filter(({operation}) => operation != null)
+      .map(({error}) => error);
+
+    if (errors.length === 0) {
+      // we couldn't find an error in a valid fixture, fallback on all errors
+      errors = fixtureOperations.map(({error}) => error);
+    }
+
+    if (errors.every((error) => error instanceof MissingOperationError)) {
+      // if we have a bunch of missing operation errors, we can gracefully
```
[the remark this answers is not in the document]
patsissons: I refactored this to something a bit cleaner still, essentially a function that merges all missing errors together. This should now cover the case where there are missing and non-missing errors at once (which will probably never happen but who knows).

packages/graphql-validate-fixtures/test/fixtures/ambiguous-operation/fixtures/projectA/another-another-fixture.json
patsissons: gave this some thought overnight, I will remove the additional tests in these fixtures because they are really covered in a different test. That should reduce their complexity down quite a bit.

packages/graphql-validate-fixtures/README.md
```diff
@@ -23,13 +23,76 @@ In order to associate a fixture with a GraphQL query or mutation in your app, yo
...
+    "bar": {
+      "schemaPath": "build/schema/bar.json",
+      "includes": "app/bar/**/*.graphql",
+      "extensions": {
+        "schemaTypesPath": "app/bar/types/graphql.ts"
+      }
+    }
+  }
+}
+```
```
lemonmade: Do we need all this stuff in this README? Seems like it should be in one place (probably `GraphQL-tool-utilities` and every consuming package just makes a passing mention and links to the full details?

packages/graphql-validate-fixtures/src/index.ts
```diff
@@ -1,68 +1,181 @@
...
+}
+
+interface FixtureOperationError {
+  ast: AST;
+  error: any;
+  operation?: Operation;
+  operationName?: string;
+}
+
+function evaluateFixture(
```
lemonmade: This flow feels really roundabout for me, or maybe it is just not sufficiently broken down into functions. I would expect it to be more something like this (pseudocode):

```ts
const projectsWithMatchingOperation = findMatchingOperations(fixture, projects);

if (projects.length === 0) {
  return noMatchingOperationsError(fixture); // if we want, we can do a fancy thing here to detect operations that had a close name, but didn't match. Might be overkill, I think for now just say we couldn't find any matching ones is enough
} if (projects.length > 1) {
  return multipleMatchingOperationsError(fixture, projectsWithMatchingOperation); // Tell them how to resolve
}

return validateAgainstProject(fixture, projectsWithMatchingOperation[0]);
```

packages/graphql-validate-fixtures/src/index.ts
```diff
@@ -1,68 +1,181 @@
...
+  return validateFixtureAgainstAST(fixture, ast);
+}
+
+function mergeMissingOperationErrors(errors: any[]) {
+  if (errors.length <= 1) {
+    return errors;
   }
 
-  const sources = await Promise.all(
+  const {otherErrors, operationNames} = errors.reduce<{
```
lemonmade: IMO this is a weird way of doing it, operation on errors instead of projects.

packages/graphql-validate-fixtures/src/index.ts
```diff
@@ -75,11 +188,24 @@ export async function evaluateFixtures(
...
 
-  return runForEachFixture(fixturePaths, (fixture) =>
-    validateFixtureAgainstAST(fixture, ast),
-  );
+  let schema: GraphQLSchema;
+
+  try {
+    schema = projectConfig.getSchema();
+  } catch (error) {
+    throw new Error(
```
lemonmade: If we get to this point, I would not try to merge it in with other "missing operation" errors — I'd rather just do a full stop and emit an error.

packages/graphql-validate-fixtures/src/index.ts
patsissons: I'll take another pass at it to clear up the flow.

packages/graphql-validate-fixtures/src/index.ts
patsissons: this is already a bit of a strange case. The `errors` have already been computed and we now just want to post-process the list. We don't have any typing information at this point so it's literally an array of zero or more things. We want to squeeze the errors list together and merge any `MissingOperationError`'s together, but leave other errors alone. The end result is an array of zero or more other errors, and zero or one Error representing the merge of `MissingOperationError`'s. I couldn't think of how to simplify this further, `reduce` was the best I could come up with keeping a mutating list of other errors and missing operation names.

packages/graphql-validate-fixtures/src/index.ts
patsissons: I believe that is the case, these errors during the operation discovery phase should all be going up the stack. the validation error processing should only happen in the evaluation phase. I will double check to confirm this.

packages/graphql-validate-fixtures/README.md
lemonmade: The utilities package is published too, IMO it is not a problem at all to just link to its docs on configuration for the project.

patsissons: ok, i'll refactor this into the utilities package then.

packages/graphql-validate-fixtures/src/index.ts
patsissons: [confirmed](https://github.com/Shopify/graphql-tools-web/pull/21/files#diff-1acdbb166626467a9e66c1c258aee06eR129). we do not handle these types of errors, they pop all the way out to the entrypoint.

packages/graphql-validate-fixtures/src/index.ts
patsissons: I just finished a pretty crazy rewrite to address this. You probably want to have another look again tomorrow. The flow is much more aligned with what you had in mind. There may be some more tweaks to make yet, but i think we're much closer now.

packages/graphql-validate-fixtures/src/index.ts
```diff
@@ -1,68 +1,66 @@
...
+
+async function getOperationsForProject(
+  projectConfig: GraphQLProjectConfig,
+): Promise<GraphQLProjectOperations> {
+  const operationPaths = projectConfig.includes
+    .map((filePath) => getGraphQLFilePath(projectConfig, filePath))
+    .reduce<string[]>((filePaths, pattern) => {
+      return filePaths.concat(glob.sync(pattern));
+    }, [])
+    .filter((operationPath) => projectConfig.includesFile(operationPath));
```
lemonmade: Why do you need this filter?

packages/graphql-validate-fixtures/src/validate.ts
```diff
@@ -37,106 +82,106 @@ export interface Validation {
...
+  ).filter(
+    (operationName): operationName is string =>
+      typeof operationName === 'string',
+  );
+}
 
-      if (key === OPERATION_MARKER) {
-        return allErrors;
-      }
+export interface FoundOperation {
```
lemonmade: Can you please pull this to the top of the file?

lemonmade: Also, this type confuses me; it has a single operation but also a bunch of operations?

packages/graphql-validate-fixtures/src/validate.ts
```diff
@@ -27,6 +27,51 @@ export interface Error {
   message: string;
 }
 
+export interface GraphQLProjectOperations {
```
lemonmade: I think this name is confusing, is this not just `GraphQLProject`?

packages/graphql-validate-fixtures/README.md
patsissons: https://github.com/Shopify/graphql-tools-web/pull/29 covers the main refactoring of the documentation

packages/graphql-validate-fixtures/src/index.ts
patsissons: This needs to get replaced with proper globbing like i am doing with #33 
The original intent was to remove excluded files from the list since we were just globbing on all includes.

packages/graphql-validate-fixtures/src/validate.ts
patsissons: I'll do some renaming to clean up this and the other `FoundOperation`. I agree they are poorly named, I think I was mentally unable to come up with a good name at the time so i just picked something quick to move forward.

PR #22, opened 2018-09-20
upgrading apollo-codegen to 0.20.2

patsissons:
our tooling is using `graphql@^0.13.0` and `apollo-codegen@^0.19.0` is using `graphql@^0.12.0`.

I am extending the built-in legacy types from `apollo-codegen` with some type overrides to improve the shape in some areas where the official types are lacking. I also had to patch a few places that were expecting slightly incorrect types and also forward propagating our internal types on collections to simplify type inference for consumers.

I had to add a new polyfill for `Array.flatmap` to support the `transformSelectionSetToLegacyIR` implementation.

I reduced the typing of `operationType` from a literal string union to just `string` (aligned with the built-in types) as we don't actually do any checking against this type anyways.

### DIFF
--- a/packages/graphql-tool-utilities/ast.ts
+++ b/packages/graphql-tool-utilities/ast.ts
@@ -1,17 +1,15 @@
 import './polyfills';
+import {DocumentNode, GraphQLInputType, GraphQLSchema} from 'graphql';
 import {
-  GraphQLSchema,
-  DocumentNode,
-  GraphQLType,
-  GraphQLOutputType,
-  GraphQLInputType,
-  GraphQLInterfaceType,
-  GraphQLObjectType,
-} from 'graphql';
-
-const {
-  compileToLegacyIR: compileToIR,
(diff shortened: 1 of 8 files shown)


packages/graphql-tool-utilities/ast.ts
```diff
@@ -21,66 +19,57 @@ export enum OperationType {
 
 export interface Variable {
   name: string;
-  type?: GraphQLInputType;
-}
-
-export interface TypedVariable {
-  name: string;
-  type: GraphQLInputType;
+  type?: GraphQLType;
```
lemonmade: Hmm this change seems incorrect? Variables can only be input types

patsissons: it's as per the internal types, i can try and change though to see if it makes a big difference. I'm guessing that will break compat with internal types though.

patsissons: I was thinking of the output type downcast, this one seems fine 👍

PR #23, opened 2018-09-26
refactoring graphql functions

patsissons:
consolidating `graphql-config` functions from `graphql-typescript-definitions` into `graphql-tool-utilities`.  This will be in preparation for #21 to start consuming common `graphql-config` related functions.

### DIFF
--- a/packages/graphql-tool-utilities/augmentations.ts
+++ b/packages/graphql-tool-utilities/augmentations.ts
@@ -1,0 +1,16 @@
+import {GraphQLProjectConfig} from 'graphql-config/lib/GraphQLProjectConfig';
+
+declare module 'graphql-config/lib/GraphQLProjectConfig' {
+  interface GraphQLProjectConfig {
+    resolveProjectName(defaultName?: string): string;
+  }
+}
+
+function resolveProjectName(
+  this: GraphQLProjectConfig,
+  defaultName = 'GraphQL',
+) {
+  return this.projectName || defaultName;
+}
+
+GraphQLProjectConfig.prototype.resolveProjectName = resolveProjectName;
--- a/packages/graphql-typescript-definitions/src/index.ts
+++ b/packages/graphql-typescript-definitions/src/index.ts
@@ -10,7 +10,7 @@ import {
   concatAST,
 } from 'graphql';
 import chalk from 'chalk';
-import {dirname, isAbsolute, resolve} from 'path';
+import {dirname, resolve} from 'path';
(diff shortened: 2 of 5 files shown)


packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -318,7 +324,7 @@ export class Builder extends EventEmitter {
       throw new Error(`No project found for file: ${filePath}`);
     }
 
-    const projectName = project.projectName || '';
+    const projectName = project.resolveProjectName('');
```
lemonmade: Do you still need the empty string argument?

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -340,38 +346,30 @@ export class Builder extends EventEmitter {
...
   }
 
   private removeDocumentForFile(filePath: string) {
     const project = this.config.getConfigForFile(filePath);
 
     if (project) {
-      const projectName = project.projectName || '';
-      const documents = this.documentMapByProject.get(projectName);
+      const documents = this.documentMapByProject.get(
+        project.resolveProjectName(''),
```
lemonmade: Same as above

packages/graphql-typescript-definitions/src/index.ts
patsissons: yup, the empty string is necessary since we want the `projectName` to be falsy when we evaluate how to construct the schema types file name.

packages/graphql-tool-utilities/augmentations.ts
```diff
@@ -0,0 +1,16 @@
...
+
+declare module 'graphql-config/lib/GraphQLProjectConfig' {
+  interface GraphQLProjectConfig {
+    resolveProjectName(defaultName?: string): string;
+  }
+}
+
+function resolveProjectName(
+  this: GraphQLProjectConfig,
+  defaultName = 'GraphQL',
```
patsissons: This is our new default project name. I am happy to change this to just about anything if anyone has opinions. 😄

packages/graphql-typescript-definitions/src/index.ts
lemonmade: No one but you is ever going to remember why this is the case, and even you'll forget soon enough. We either need to come up with a more elegant solution or this needs a bunch of comments.

patsissons: it looks like `Map<K, V>` supports `undefined` as a key, so i don't need to do any of this non-sense.

PR #25, opened 2018-09-27
Fixes non-unique List responses

cartogram:
Items of a list type were returning the same data, which does not properly represent fake a real response. This PR looks for List types and adds the index of the field to the end of the `responseName` in order to force a different value for each field. I am open to a better ways to achieve this.

### DIFF
--- a/packages/graphql-fixtures/src/fill.ts
+++ b/packages/graphql-fixtures/src/fill.ts
@@ -21,10 +21,13 @@ import {
 } from 'graphql-tool-utilities';
 import {randomFromArray, chooseNull} from './utilities';
 
-export type FieldDetails = (Field | InlineFragment) & {
+export interface FieldMetadata {
+  fieldIndex?: number;
   fieldName: string;
   responseName: string;
-};
+}
+
+export type FieldDetails = (Field | InlineFragment) & FieldMetadata;
 
 export interface ResolveDetails {
   type: GraphQLType;
@@ -198,8 +201,21 @@ function unwrapThunk<T>(value: Thunk<T>, details: ResolveDetails): T {
   return isResolver(value) ? value({...details, type: unwrappedType}) : value;
 }
 
-function withRandom<T>(keypath: FieldDetails[], func: () => T) {
-  faker.seed(seedFromKeypath(keypath.map(({responseName}) => responseName)));
+function keyPathElement(responseName: string, fieldIndex: number | undefined) {
+  return fieldIndex == null ? responseName : `${responseName}[${fieldIndex}]`;
+}
+
+// we need to set a seedOffset when filling fields that are indexed leafs in the
+// graph, for indexed objects in the graph their key path will use the parent
+// field index instead.
+function withRandom<T>(keypath: FieldDetails[], func: () => T, seedOffset = 0) {
+  faker.seed(
+    seedFromKeypath(
+      keypath.map(({fieldIndex, responseName}) =>
+        keyPathElement(responseName, fieldIndex),
+      ),
+    ) + seedOffset,
+  );
   const value = func();
   faker.seed(Math.random() * 10000);
   return value;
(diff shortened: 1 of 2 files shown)


packages/graphql-fixtures/src/fill.ts
```diff
@@ -264,7 +264,10 @@ function fillType(
       ? array.map((value: any) =>
           fillType(
             unwrappedType.ofType,
-            field,
+            {
+              ...field,
+              responseName: (field.responseName += parentFields.indexOf(field)),
```
cartogram: This is only actually appending a `1`. If the general approach is correct here, I'll generate something unique here.

lemonmade: you should just be able to use the index from the map here

lemonmade: I don't think you want to mutate the `responseName` though, that might have other consequences (for example, the filler for a field gets the details about the field, including its response name, and is probably not expecting it to differ from the original value. I think we may need another argument for this.

patsissons: can we not just add a new `fieldIndex` to `FieldDetails` (defaulting to `undefined`) and refactor how we build the key to consider the existence of this field index?

packages/graphql-fixtures/src/fill.ts
```diff
@@ -198,8 +201,23 @@ function unwrapThunk<T>(value: Thunk<T>, details: ResolveDetails): T {
...
+function withRandom<T>(
+  // {fieldIndex = 0}: FieldDetails,
+  keypath: FieldDetails[],
+  func: () => T,
+  fieldIndex = 0,
+) {
+  faker.seed(
+    seedFromKeypath(
+      keypath.map(({fieldIndex, responseName}) =>
+        keyPathElement(responseName, fieldIndex),
```
cartogram: Rename `fieldIndex` to prevent the shodowing.

packages/graphql-fixtures/src/fill.ts
```diff
@@ -198,8 +201,23 @@ function unwrapThunk<T>(value: Thunk<T>, details: ResolveDetails): T {
...
+      ),
+    ) + fieldIndex,
```
cartogram: Is this still needed?

patsissons: i believe so. there should be two cases, one where one of our ancestors is indexed and another where the field is indexed as a leaf in the graph. This `+ fieldIndex` covers the leaf scenario. The alternative is to inject the field being filled into the keypath, but this just seemed a bit easier at first glance. the other method (essentially `withRandom([...details.parentFields, details])`) is probably just as viable but slightly less performant.

packages/graphql-fixtures/src/fill.ts
```diff
@@ -198,8 +201,21 @@ function unwrapThunk<T>(value: Thunk<T>, details: ResolveDetails): T {
...
-function withRandom<T>(keypath: FieldDetails[], func: () => T) {
-  faker.seed(seedFromKeypath(keypath.map(({responseName}) => responseName)));
+function keyPathElement(responseName: string, fieldIndex: number | undefined) {
+  return fieldIndex == null ? responseName : `${responseName}[${fieldIndex}]`;
+}
+
+// we need to set a seedOffset when filling fields that are indexed leafs in the
+// graph, for indexed objects in the graph their key path will use the parent
+// field index instead.
+function withRandom<T>(keypath: FieldDetails[], func: () => T, seedOffset = 0) {
```
lemonmade: I am confused, why do you need both this offset and the index? I guess I don't understand what the comment above is trying to tell me.

patsissons: `seedOffset` is for when the field we're filling is itself indexed (i.e., `details` from `createValue`) because otherwise it just relies on the parent key path for seeding which would be the same for all of its peer nodes.

patsissons:
to be re-visited in #50

PR #26, opened 2018-09-27
adding support for vscode-jest

patsissons:
* updating `settings.json` so that `jest` uses our `config.json`
* adding `launch.json` config so that `vscode-jest` supports debugger integration
* ading `extensions.json` to recommend `vscode-jest` (as well as `eslint` and `prettier` extensions)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,22 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,22 @@
...
+        "--runInBand",
+        "--config",
+        "${workspaceFolder}/config/jest/config.json"
+      ],
+      "cwd": "${workspaceFolder}",
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen"
+    }
+  ]
+}
```
lemonmade: This is separate from the vscode-jest stuff, no?

patsissons: the extension actually uses a specifically named launch config to enable custom debug arguments (required for our `ts-jest` and `config.json`)

PR #27, opened 2018-10-01
fixing dependencies for graphql-tools-utilities

patsissons:
`apollo-codegen` is a monorepo that pulls in a lot of extra dependencies that are not necessary, so we can reduce this to just `apollo-codegen-core`.

additionally, consolidating which version of the graphql tooling we are using to simplify versioning in versions in consumers. This should essentially be a no-op change that will only remove no longer necessary packages from the dependency chain.

* `@types/graphql` is upgraded to `0.13.4`
* packages in this repo no longer directly depend on `@types/graphql`, `graphql`, or `graphql-config` (now provided through `graphql-tool-utilities`)

### DIFF
--- a/packages/graphql-tool-utilities/package.json
+++ b/packages/graphql-tool-utilities/package.json
@@ -30,8 +30,9 @@
   },
   "dependencies": {
     "@types/graphql": "^0.13.0",
-    "apollo-codegen": "^0.20.0",
+    "apollo-codegen-core": "0.20.1",
     "core-js": "^2.4.1",
-    "graphql-config": "^2.0.0"
+    "graphql": "0.13.2",
+    "graphql-config": "2.1.0"
   }
 }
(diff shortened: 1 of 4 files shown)


packages/graphql-fixtures/package.json
```diff
@@ -26,7 +26,7 @@
   "dependencies": {
     "@types/faker": "^4.1.2",
     "faker": "^4.1.0",
-    "graphql-tool-utilities": "^0.7.4"
+    "graphql-tool-utilities": "0.7.4"
```
lemonmade: You are not supposed to change this, lerna should be managing these dependencies

patsissons: how does this magic dependency management work?

patsissons: ok, restored the original version specs (and learned how to use lerna!)

PR #30, opened 2018-10-03
refactoring file watcher instrumentation

### LINKED ISSUE #28: duplicate calls to build graphql type definitions on startup
When running the `graphql-typescript-definitions`, the `chokidar` file system watcher will by default automatically emit `add` events for all discovered files on the initial scans. Due to timing issues, this may not have been observed until recently. This initial set of events can be omitted if `chokidar` is given a [`ignoreInitial` option](https://github.com/paulmillr/chokidar#path-filtering).

### LINKED ISSUE #31: graphql-typescript-definitions does not ignore excludes
globbing only includes patterns defined in the `includes` configuration, this results in graphql files being processed that specifically should not be processed. In a workspace where all graphql files are included in some project this has no effect, but for any workspace with graphql files that should be fully excluded from processing this can cause errors if an `include` glob pattern matches an excluded file. This can be particularly visible in a single project workspace as any excluded files would not be linked to any project.

cc @keyfer

patsissons:
resolves #28 

Setting the `ignoreInitial` option when we setup `chokidar` watchers to prevent file `add` events from being emitted on startup which can cause duplicate calls to the `generateDocumentTypes` function.

watchers are now composed per-project and one additional watcher for all schema files. Watchers and globbing now use both `includes` and `excludes` when constructing file matchers. This should eliminate any issues where excluded files are being included from an `include` in another project. This also enables a much cleaner code path for processing file paths because we should always have the current project config for the file path in scope.

This will likely also resolve #31

### DIFF
--- a/packages/graphql-typescript-definitions/src/index.ts
+++ b/packages/graphql-typescript-definitions/src/index.ts
@@ -8,9 +8,9 @@ import {
   concatAST,
 } from 'graphql';
 import chalk from 'chalk';
-import {dirname, resolve} from 'path';
+import {dirname, join, resolve} from 'path';
 import {readFile, writeFile, mkdirp} from 'fs-extra';
-import {watch} from 'chokidar';
+import {FSWatcher, watch} from 'chokidar';
 import * as glob from 'glob';
 import {
   getGraphQLConfig,
@@ -78,6 +78,7 @@ export class Builder extends EventEmitter {
     string | undefined,
     Map<string, DocumentNode>
   >();
+  private readonly watchers: FSWatcher[] = [];
 
   constructor({cwd, ...options}: BuilderOptions) {
     super();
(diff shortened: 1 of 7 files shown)


patsissons:
going to try and get a test written to cover this case.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -189,6 +164,80 @@ export class Builder extends EventEmitter {
...
+        ),
+        {
+          ignored: projectConfig.excludes.map((exclude) =>
+            getGraphQLFilePath(this.config, exclude),
+          ),
+          ignoreInitial: false,
+        },
+      )
+        .on('add', update) // TODO: async function not supported
+        .on('change', update) // TODO: async function not supported
```
patsissons: this here is problematic because `chokidar` does not support async event handling. so we are leaking promises like crazy on `change` and `add` events.

patsissons:
This has been replaced by #33

PR #33, opened 2018-10-03
using graphql-config for all schema file handling

### LINKED ISSUE #5: Add support for `graphql-config`
https://github.com/prismagraphql/graphql-config

### LINKED ISSUE #28: duplicate calls to build graphql type definitions on startup

### LINKED ISSUE #31: graphql-typescript-definitions does not ignore excludes

### LINKED ISSUE #32: graphql-typescript-definitions does not support graphql schema files
when generating types we also sometimes write out the schema file. This only makes sense if our configured `schemaPath` is a JSON file. If our configured `schemaPath` is a graphql file then we should omit this step.

patsissons:
This PR removes what was left of legacy graphql file reading in `graphql-typescript-definitions` to process changes to the graphql schema file. We now rely entirely on the `graphql-config` instance by using the `getSchema()` function. This resolves #32

The `chokidar` file watchers and the `glob` function now respect `excludes` defined in the `graphql-config` file. Each project now produces its own pair of `chokidar` watchers (one for schema and one for documents). Similarly, globbing now occurs on a per-project basis. This allows projects to specifically cover only the files they want linked to a schema. This resolves #31 

`chokidar` no longer generates events on startup. Setting the `ignoreInitial` option when we setup `chokidar` watchers prevents file add events from being emitted on startup which were causing duplicate calls to the `generateDocumentTypes` function. This resolves #28 

This will end up being the last piece of the `graphql-config` puzzle. This resolves #5 

replaces #30
blocked on #37 and #38

### DIFF
--- a/packages/graphql-typescript-definitions/src/index.ts
+++ b/packages/graphql-typescript-definitions/src/index.ts
@@ -1,7 +1,5 @@
 import {EventEmitter} from 'events';
 import {
-  buildClientSchema,
-  GraphQLSchema,
   DocumentNode,
   DefinitionNode,
   OperationDefinitionNode,
@@ -11,15 +9,15 @@ import {
 } from 'graphql';
 import chalk from 'chalk';
 import {dirname, join, resolve} from 'path';
-import {readJSON, readFile, writeFile, mkdirp} from 'fs-extra';
-import {watch} from 'chokidar';
-import * as glob from 'glob';
+import {mkdirp, readFile, writeFile} from 'fs-extra';
+import {FSWatcher, watch} from 'chokidar';
 import {
   getGraphQLConfig,
   GraphQLProjectConfig,
   GraphQLConfig,
 } from 'graphql-config';
 import {
+  getGraphQLProjectIncludedFilePaths,
   getGraphQLProjectForSchemaPath,
   getGraphQLProjects,
   getGraphQLSchemaPaths,
@@ -79,6 +77,7 @@ export class Builder extends EventEmitter {
     string | undefined,
     Map<string, DocumentNode>
   >();
+  private readonly watchers: FSWatcher[] = [];
 
   constructor({cwd, ...options}: BuilderOptions) {
     super();
@@ -121,7 +120,6 @@ export class Builder extends EventEmitter {
   }
 
   async run({watch: watchGlobs = false} = {}) {
-    const globs = this.getGlobs();
     let schemaPaths: string[];
 
     try {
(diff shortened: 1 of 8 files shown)


patsissons:
I think i will merge these changes into #33 as it will essentially cover the majority of what this PR is already trying to accomplish.

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -350,48 +494,37 @@ export class Builder extends EventEmitter {
...
-
-    if (project) {
-      const documents = this.documentMapByProject.get(project.projectName);
-
-      if (documents) {
-        documents.delete(filePath);
-      }
-    }
-  }
+function getUniqueFilePaths(filePaths: string[]) {
```
patsissons: I don't think this is necessary, there should be no duplicate file paths unless we don't trust `glob`

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -1,7 +1,5 @@
 import {EventEmitter} from 'events';
```
lemonmade: I still see one reference in the README to passing an explicit list of globs to the CLI — we should get rid of that, right?

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -189,14 +171,95 @@ export class Builder extends EventEmitter {
...
+      watcher.close();
+    });
+
+    this.watchers.length = 0;
+  }
+
+  private setupDocumentWatchers() {
+    const update = (filePath: string, projectConfig: GraphQLProjectConfig) => {
+      try {
+        this.updateDocumentForFileSync(filePath, projectConfig);
```
lemonmade: So, I am confused by this change. You reference #28 in your PR description as to why we move to sync functions for these callbacks, but it seems that just the `ignoreInitial` change addresses that. I worry a lot about doing all of the synchronous operations since this whole thing happens alongside other expensive operations in Sewing Kit.

patsissons: the `ignoreInitial` does fix the specific problem that #28 represents, but it does not fix the fact that we cannot _safely_ run async code in the `chokidar` event handlers (The reference to #28 was just that part of that effort resolves it, the remainder of the effort resolves the async issues). These sync calls in the `update` function ought to be limited to actual file changes, so in general we should not be hitting this function too often. All of the _spin up_ processing should be handled by the async `run` function which is offloading all processing onto async functions.

packages/graphql-typescript-definitions/test/__snapshots__/cli.test.ts.snap
```diff
@@ -71,16 +71,18 @@ Object {
   },
   "stderr": "",
   "stdout": "
- ERROR  Error parsing 'packages/graphql-typescript-definitions/test/fixtures/missing-schema/schema.json':
-
-ENOENT: no such file or directory, open 'packages/graphql-typescript-definitions/test/fixtures/missing-schema/schema.json'
-Error: Error parsing 'packages/graphql-typescript-definitions/test/fixtures/missing-schema/schema.json':
-
-ENOENT: no such file or directory, open 'packages/graphql-typescript-definitions/test/fixtures/missing-schema/schema.json'
+ ERROR  ENOENT: no such file or directory, open 'packages/graphql-typescript-definitions/test/fixtures/missing-schema/schema.json'
```
patsissons: I think I'm going to try and repair this in another PR to restore the original error handling.

patsissons: resolved in #37

PR #34, opened 2018-10-05
adding augmentation to simplify project schemaPath resolution

patsissons:
`GraphQLProjectConfig.schemaPath` is [nullable](https://github.com/prisma/graphql-config/blob/master/src/GraphQLProjectConfig.ts#L92) because it's possible to [share a schema](https://github.com/prisma/graphql-config/blob/4c09eef3b97ebfbde6aa3710dfcfb1fe845f6d60/specification.md#default-configuration-properties) across multiple projects. Due to the typing used in `graphql-config` this results in the `schemaPath` having a nullable type. In practice, a project without a schema is not valid. In order to reduce boilerplate to guard against an invalid schema, a resolution function is added to the project to simplify _access or throw_ on the `schemaPath`.

### DIFF
--- a/packages/graphql-tool-utilities/augmentations.ts
+++ b/packages/graphql-tool-utilities/augmentations.ts
@@ -3,6 +3,7 @@ import {GraphQLProjectConfig} from 'graphql-config/lib/GraphQLProjectConfig';
 declare module 'graphql-config/lib/GraphQLProjectConfig' {
   interface GraphQLProjectConfig {
     resolveProjectName(defaultName?: string): string;
+    resolveSchemaPath(): string;
   }
 }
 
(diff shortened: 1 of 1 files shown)


packages/graphql-tool-utilities/augmentations.ts
```diff
@@ -3,6 +3,7 @@ import {GraphQLProjectConfig} from 'graphql-config/lib/GraphQLProjectConfig';
 declare module 'graphql-config/lib/GraphQLProjectConfig' {
   interface GraphQLProjectConfig {
     resolveProjectName(defaultName?: string): string;
+    resolveSchemaPath(): string;
```
lemonmade: Any reason not to do this as a getter?

patsissons: no particular reason, probably just to be consistent with `resolveProjectName`.

PR #36, opened 2018-10-09
adding getGraphQLProjectIncludedFilePaths function

patsissons:
Adding new utility function to unify how to generate a list of all included file paths for a specific `GraphQLProjectConfig` instance. Refactoring usage in `graphql-validate-fixtures` to use this new function. #33 will handle the refactoring for `graphql-typescript-definitions`.

### DIFF
--- a/packages/graphql-tool-utilities/config.ts
+++ b/packages/graphql-tool-utilities/config.ts
@@ -1,5 +1,7 @@
+import glob from 'glob';
 import {GraphQLConfig, GraphQLProjectConfig} from 'graphql-config';
 import {isAbsolute, resolve} from 'path';
+import {promisify} from 'util';
 
 import './augmentations';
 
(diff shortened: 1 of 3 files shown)


packages/graphql-tool-utilities/config.ts
```diff
@@ -30,6 +32,20 @@ export function getGraphQLSchemaPaths(config: GraphQLConfig) {
...
+) {
+  return (await Promise.all(
+    projectConfig.includes.map((include) =>
+      promisify(glob)(getGraphQLFilePath(projectConfig, include), {
+        ignore: projectConfig.excludes.map((exclude) =>
+          getGraphQLFilePath(projectConfig, exclude),
+        ),
+      }),
+    ),
+  )).flatMap((filePaths) => filePaths);
```
lemonmade: This is available in our version of Node?!

patsissons: good question. it's actually getting polyfilled by `apollo-codegen-core`. Should we maybe poly that in more properly?

patsissons: correction. `core-js/es7/array` is pulling in the implementation.

patsissons: I think I should have added `esnext.array` to `tsconfig.json` libs a while back. `apollo-codegen-core` just happens to also include a `flatMap` type def.

PR #37, opened 2018-10-09
unifying error handling of missing schema files

patsissons:
Now that the tooling is relying completely on `graphql-config` for resolution of graphql entities, we are also at the mercy of `graphql-config` for error handling (which it performs less than stellar at). The `resolveSchemaPath` function should be the primary entrypoint for all consumers to resolve the `schemaPath` of a `GraphQLProjectConfig` instance, therefore we can additionally check the existence of the schema file and throw a more human readable error (instead of relying on node's readFile error, `ENOENT: ...`)

### 🎩 Notes

I went with `existsSync` to prevent requiring an additional `fs-extra` dependency and because it didn't force consumers to use async functions. Open to change suggestions.

### DIFF
--- a/packages/graphql-tool-utilities/config.ts
+++ b/packages/graphql-tool-utilities/config.ts
@@ -26,9 +26,7 @@ export function getGraphQLProjects(config: GraphQLConfig) {
 
 export function getGraphQLSchemaPaths(config: GraphQLConfig) {
   return getGraphQLProjects(config).reduce<string[]>((schemas, project) => {
-    return schemas.concat(
-      getGraphQLFilePath(config, project.resolveSchemaPath()),
-    );
+    return schemas.concat(project.resolveSchemaPath());
   }, []);
 }
 
--- a/packages/graphql-typescript-definitions/src/index.ts
+++ b/packages/graphql-typescript-definitions/src/index.ts
@@ -123,7 +123,14 @@ export class Builder extends EventEmitter {
 
   async run({watch: watchGlobs = false} = {}) {
     const globs = this.getGlobs();
-    const schemaPaths = this.getSchemaPaths();
+    let schemaPaths: string[];
+
+    try {
+      schemaPaths = getGraphQLSchemaPaths(this.config);
+    } catch (error) {
+      this.emit('error', error);
+      return;
+    }
 
     const update = async (filePath: string) => {
       try {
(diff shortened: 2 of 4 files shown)


packages/graphql-tool-utilities/config.ts
```diff
@@ -26,9 +26,7 @@ export function getGraphQLProjects(config: GraphQLConfig) {
 
 export function getGraphQLSchemaPaths(config: GraphQLConfig) {
   return getGraphQLProjects(config).reduce<string[]>((schemas, project) => {
-    return schemas.concat(
-      getGraphQLFilePath(config, project.resolveSchemaPath()),
-    );
+    return schemas.concat(project.resolveSchemaPath());
```
patsissons: `resolveSchemaPath` now performs the full path resolution, so we don't need to use `getGraphQLFilePath` any longer (i will eliminate `getGraphQLFilePath` in a new PR shortly).

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -123,7 +123,14 @@ export class Builder extends EventEmitter {
 
   async run({watch: watchGlobs = false} = {}) {
     const globs = this.getGlobs();
-    const schemaPaths = this.getSchemaPaths();
+    let schemaPaths: string[];
+
+    try {
+      schemaPaths = getGraphQLSchemaPaths(this.config);
+    } catch (error) {
+      this.emit('error', error);
```
patsissons: emit our new error and exit early since we don't have a stable graphql configuration to continue processing.

packages/graphql-typescript-definitions/test/__snapshots__/cli.test.ts.snap
```diff
@@ -70,17 +70,18 @@ Object {
...
   },
   "stderr": "",
-  "stdout": "
- ERROR  Error parsing 'packages/graphql-typescript-definitions/test/fixtures/missing-schema/schema.json':
-
-ENOENT: no such file or directory, open 'packages/graphql-typescript-definitions/test/fixtures/missing-schema/schema.json'
-Error: Error parsing 'packages/graphql-typescript-definitions/test/fixtures/missing-schema/schema.json':
-
-ENOENT: no such file or directory, open 'packages/graphql-typescript-definitions/test/fixtures/missing-schema/schema.json'
+  "stdout": " ERROR  schemaPath not found for project 'GraphQL': 'packages/graphql-typescript-definitions/test/fixtures/missing-schema/schema.json'
```
lemonmade: Maybe instead of just printing the path, say “expected to find it in X based on GraphQL config file, but no such file existed.”

packages/graphql-tool-utilities/config.ts
patsissons: #38

PR #38, opened 2018-10-09
Remove getGraphQLFilePath function

patsissons:
This function was added to simplify fully qualified resolution of paths relative to the `graphql-config` file. `GraphQLProjectConfig` already exposes this functionality wtih `resolveConfigPath`. This PR is refactoring usages of `getGraphQLFilePath` with `resolveConfigPath`.

this will conflict slightly with #37 and can be resolved by taking the change in #37

### DIFF
--- a/packages/graphql-tool-utilities/config.ts
+++ b/packages/graphql-tool-utilities/config.ts
@@ -1,6 +1,5 @@
 import glob from 'glob';
 import {GraphQLConfig, GraphQLProjectConfig} from 'graphql-config';
-import {isAbsolute, resolve} from 'path';
 import {promisify} from 'util';
 
 import './augmentations';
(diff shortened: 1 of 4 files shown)


packages/graphql-tool-utilities/config.ts
```diff
@@ -37,9 +36,9 @@ export async function getGraphQLProjectIncludedFilePaths(
 ) {
   return (await Promise.all(
     projectConfig.includes.map((include) =>
-      promisify(glob)(getGraphQLFilePath(projectConfig, include), {
+      promisify(glob)(projectConfig.resolveConfigPath(include), {
         ignore: projectConfig.excludes.map((exclude) =>
-          getGraphQLFilePath(projectConfig, exclude),
+          projectConfig.resolveConfigPath(exclude),
```
lemonmade: Hmm this name seems misleading since it does not resolve the config path, it resolves the path relative to the config, no?

patsissons: it's true, the `graphql-config` function name is a bit misleading. but short of adding an augmentation to fix the misleading function name there isn't much to be done. I'm hoping that this is to be rarely used outside of this tooling. That said, anyone using `graphql-config` will have access to `resolveConfigPath` so it's probably not worth fighting too hard. I'll get a PR up in `graphql-config` to refactor the name.

patsissons: I posted a PR but I'm not sure what kind of turn around time to expect.

lemonmade: Can we not just add this as a separate method on the projects we return? Since we wrap all the graphql config stuff seems like that should be doable.

patsissons:
@lemonmade can i get a ✅ on this one so that i can base #33 on top and patch with the new `resolvePathRelativeToConfig` function.

PR #43, opened 2018-10-10
adding option to opt out of schemaPath file existence validation

patsissons:
adds an optional `mustExist` parameter to `resolveSchemaPath` to opt out the file system check if the file exists. This is useful for consumers of this function who intend to create or delete the file and don't necessarily care if it exists yet, but rather just want to know where it should exist.


patsissons:
going with `ignoreMissing` 😄

PR #46, opened 2018-10-15
updating yarn and yarn.lock to latest 1.10.1

patsissons:
Changed yarn minimum version to `>=1.10` and added hash to `yarn.lock`


patsissons:
should those lock files even exist? they were hidden via vscode settings. I don't think they are necessary unless we were using submodules and the packages could be worked on individually without the top level lock file? In any case, none have been updated in ages, so let's try getting rid of them.

PR #49, opened 2018-10-17
restructuring project to follow src/test/lib layout

patsissons:
includes some minor cleanup adjustments to the code in response to the added tests


packages/graphql-tool-utilities/package.json
```diff
@@ -25,9 +27,6 @@
     "pretest": "yarn build",
     "prepublishOnly": "yarn build"
   },
-  "devDependencies": {
-    "typescript": "^2.8.3"
```
patsissons: this is probably just a relic of old times. should not be necessary with the root level `devDependency`

packages/graphql-tool-utilities/src/augmentations.ts
```diff
@@ -64,6 +56,10 @@ function resolveSchemaPath(this: GraphQLProjectConfig, ignoreMissing = false) {
   return schemaPath;
 }
 
-GraphQLProjectConfig.prototype.resolvePathRelativeToConfig = resolvePathRelativeToConfig;
+// temporary augmentation until `graphql-config` supports this new function
+// see: https://github.com/prisma/graphql-config/pull/113
+GraphQLProjectConfig.prototype.resolvePathRelativeToConfig =
+  GraphQLProjectConfig.prototype.resolveConfigPath;
```
patsissons: I turned this into a real alias instead of a wrapping alias. deleted the wrapper above.

packages/graphql-tool-utilities/src/augmentations.ts
```diff
@@ -9,19 +9,11 @@ declare module 'graphql-config/lib/GraphQLProjectConfig' {
...
-// temporary augmentation until `graphql-config` supports this new function
-// see: https://github.com/prisma/graphql-config/pull/113
-function resolvePathRelativeToConfig(
-  this: GraphQLProjectConfig,
-  relativePath: string,
-) {
-  // this is just an alias to resolveConfigPath with a more meaningful name
-  return this.resolveConfigPath(relativePath);
-}
+export const defaultGraphQLProjectName = 'GraphQL';
```
patsissons: exposed this mostly for testing purposes

packages/graphql-tool-utilities/src/config.ts
```diff
@@ -2,6 +2,7 @@ import {GraphQLConfig, GraphQLProjectConfig} from 'graphql-config';
 import {promisify} from 'util';
 
 import './augmentations';
+import './polyfills';
```
patsissons: this should have been here all along. it was probably getting pulled in by the `ast` module instead which hid the fact that this module was technically broken at runtime (when in isolation).

packages/graphql-tool-utilities/src/config.ts
```diff
@@ -17,7 +18,7 @@ export function getGraphQLProjects(config: GraphQLConfig) {
 
   const project = config.getProjectConfig();
 
-  if (project) {
+  if (project && project.schemaPath) {
```
patsissons: without validation, a totally invalid project config actually returns a project instance. this check on the `schemaPath` prevents that from happening so we drop into our expected error thrown scenario.

packages/graphql-tool-utilities/src/index.ts
```diff
@@ -1,5 +1,3 @@
-import './augmentations';
```
patsissons: this is not necessary, our `config` module will pull in the augmentations.

tsconfig.json
```diff
@@ -1,5 +1,10 @@
 {
   "compilerOptions": {
+    "baseUrl": "./packages",
+    "paths": {
+      "graphql-tool-utilities/lib/*": ["./graphql-tool-utilities/src/*"],
+      "graphql-tool-utilities": ["./graphql-tool-utilities/src"]
```
patsissons: this is required to satisfy the lerna TypeScript mono-repo setup (when inter-project dependencies are involved). Based on the [official guidance](https://github.com/Quramy/lerna-yarn-workspaces-example#resolve-dependencies-as-typescript-modules), and modified slightly because we access `lib` submodules in our case.

TheMallen: Yep we did this in quilt too

packages/graphql-fixtures/src/fill.ts
```diff
@@ -18,7 +18,7 @@ import {
   Field,
   Operation,
   InlineFragment,
-} from 'graphql-tool-utilities/ast';
+} from 'graphql-tool-utilities/lib/ast';
```
TheMallen: importing from lib seems wrong?

patsissons: I'm not sure how to make this work properly otherwise, since the `ast` module will live in `graphql-tool-utilities/lib/ast` at runtime.

PR #55, opened 2018-10-31
augmentation to handle importing GraphQL SDL schemas with schema definitions

patsissons:
any imported `SDL` schema that includes a schema definition can result in empty schema types because of [a bug](https://github.com/prisma/graphql-import/issues/190) in `graphql-import`.

if we have an SDL to import that looks like

```graphql
schema {
  query: QueryRoot
}

type QueryRoot {
  foo: String
}
```

our imported schema will not include a _query type_ as defined by the `SchemaDefinition` at the top.

This PR hooks into the `readSchema` function used by `graphql-config` to load the `SDL` schema manually rather than relying on `graphql-import`.


patsissons:
I'm going to try and augment `graphql-import` instead since it's a little bit cleaner

lemonmade:
@patsissons are you still working on this?

patsissons:
I need to rewrite this PR without augmentations now that they have been deprecated.

PR #56, opened 2018-11-14
fixing issue where watch mode locks when includes is omitted

patsissons:
filtering out GraphQL projects that have no `includes` defined from initializing a chokidar watcher.

chokidar hangs on the `ready` event if there are no file globs to watch, so we must not create a watcher for any projects that do not include any paths to watch.

### DIFF
--- a/packages/graphql-typescript-definitions/src/index.ts
+++ b/packages/graphql-typescript-definitions/src/index.ts
@@ -191,32 +191,34 @@ export class Builder extends EventEmitter {
       await this.generateDocumentTypes();
     };
 
-    return getGraphQLProjects(this.config).map((projectConfig) => {
-      return watch(
-        projectConfig.includes.map((include) =>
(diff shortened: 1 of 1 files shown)


packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -191,32 +191,34 @@ export class Builder extends EventEmitter {
...
-    return getGraphQLProjects(this.config).map((projectConfig) => {
-      return watch(
-        projectConfig.includes.map((include) =>
-          projectConfig.resolvePathRelativeToConfig(include),
-        ),
-        {
-          ignored: projectConfig.excludes.map((exclude) =>
-            projectConfig.resolvePathRelativeToConfig(exclude),
+    return getGraphQLProjects(this.config)
+      .filter(({includes}) => includes.length > 0)
```
patsissons: this is the only actual change, everything else is just whitespace.

Issue #57, opened 2018-11-21
Make enums from graphql-typescript-definitions tree-shakeable


tzvipm:
Consider making the root types file tree-shakeable by making each enum its own module that gets re-exported from the root types file.

PR #61, opened 2018-11-23
[WIP] Add schema types tree shaking support

tzvipm:
Closes https://github.com/Shopify/graphql-tools-web/issues/57


packages/graphql-typescript-definitions/src/print/schema/index.ts
```diff
@@ -35,13 +39,34 @@ export function printSchema(schema: GraphQLSchema, options: Options = {}) {
...
+        generate(
+          t.file(t.program([t.exportNamedDeclaration(enumType, [])]), [], []),
+        ).code,
+      );
+      fileBody.unshift(
+        t.importDeclaration(
+          [t.importSpecifier(enumType.id, enumType.id)],
+          t.stringLiteral(`./${enumType.id.name}`),
+        ),
+        t.exportNamedDeclaration(null, [
```
lemonmade: I think we can probably `export {default as NAME} from './NAME';` here?

packages/graphql-typescript-definitions/src/print/schema/index.ts
```diff
@@ -35,13 +39,34 @@ export function printSchema(schema: GraphQLSchema, options: Options = {}) {
...
-    );
+    if (isEnumType(type)) {
+      const enumType = tsEnumForType(type, options);
+      fileMap.set(
+        `${enumType.id.name}.ts`,
+        generate(
+          t.file(t.program([t.exportNamedDeclaration(enumType, [])]), [], []),
+        ).code,
+      );
+      fileBody.unshift(
```
lemonmade: I find it a bit hard to read that half of these blocks sets the AST, the other half sets a fully rendered page. Can we just set the AST in both cases, and then do a single pass to "stringify" the contents afterwards?

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -247,9 +247,16 @@ export class Builder extends EventEmitter {
...
 
     const schemaTypesPath = getSchemaTypesPath(projectConfig, this.options);
-    const definition = printSchema(projectConfig.getSchema(), this.options);
-    await mkdirp(dirname(schemaTypesPath));
-    await writeFile(schemaTypesPath, definition);
+    const definitions = generateSchemaTypes(
+      projectConfig.getSchema(),
+      this.options,
+    );
+    await mkdirp(schemaTypesPath);
```
lemonmade: Need to make sure the documentation is updated to indicate that this MUST BE a directory now, not a file.

packages/graphql-typescript-definitions/src/print/schema/index.ts
patsissons: I don't think you can `export default` for an enum in a single statement. It's not to say we couldn't create two statements in each enum module, but it seemed unnecessary to add the complexity.

lemonmade: Yeah there should just be one line for the export (whatever that needs to be, default export makes most sense but doesn't matter much), and one for the combined import/ export in the index (more what I was getting at with my comment)

patsissons: ah for the single export in the index module. The issue there is that some types may need to reference some of the enum types, so we need to import them first.

lemonmade: Ah ok, makes sense

packages/graphql-typescript-definitions/src/print/schema/index.ts
patsissons: the primary purpose of this function is to generate code for type exports, so the file name map is the core state being computed. But we need to keep track of intermediate statements that are not enum type exports so that we can produce the _left over_ types at the end. I'm not sure if there is a better way to do this. We could certainly keep track of the enum `Statement` instances instead, but then we would have to re-iterate over everything again to produce each code fragment.

packages/graphql-typescript-definitions/test/fixtures/all-clear/.graphqlconfig
```diff
@@ -2,6 +2,6 @@
   "schemaPath": "schema.json",
   "includes": ["documents/**/*.graphql"],
   "extensions": {
-    "schemaTypesPath": "schema.ts"
+    "schemaTypesPath": "types"
```
patsissons: i'm betting you can get rid of these extensions in the fixtures, they were put in place to force the output filename to be `schema.ts` but now that it's just a directory name it is no longer necessary.

PR #64, opened 2019-01-17
Generate "empty Other" type for unions and interfaces

lemonmade:
Currently, if a union or interface has all its possible types accounted for, or if there is only a single implementing type, we do not generate the "Other" type for a union or interface. That is, given the following schema:

```graphql
interface Animal {
  legs: Int!
}

type Dog implements Animal {
  legs: Int!
}

type Person {
  name: String!
  pets: [Animal]!
}

union Creature = Person | Dog

type Query {
  creature: Creature
}
```

and this query:

```graphql
query MyQuery {
  creature {
    ... on Dog { legs }
    ... on Person {
      pets { ... on Dog { legs } }
    }
  }
}
```

We generate the following types:

```ts
namespace MyQueryData {
  type CreatureDog = {} // for creature as Dog
  type CreaturePerson = {} // for creature as Person
  type CreaturePersonPets = {} // for creature.pets
}
```

This PR changes this type generation to always create an "Other" for each of these, even if there are no other types available. This is important because, if an implementing type is added to an interface/ a new type added to a union, the user of these types has no way to protect themselves; they have no alternative type they can reference that will continue to be present after the schema change. For example, adding the following to the schema:

```graphql
type Cat implements Animal {
  legs: Int!
  lives: Int!
}

union Creature = Person | Dog | Cat
```

Would update the types to the following:

```ts
namespace MyQueryData {
  type CreatureDog = {}
  type CreaturePerson = {}
  type CreatureOther = {} // NEW
  type CreaturePersonPetsDog = {} // CHANGED: now needs to be refined to the implementing type
  type CreaturePersonPetsOther = {} // NEW
}
```

This PR simply makes it so that the types above are generated always, even if they may not technically be necessary. It also uses a `__typename: never` for them if typename is added, so that it continues to accurately reflect what data can come back.

cc/ @patsissons does this seem insane?
cc/ @loic-d since this would have helped what you ran into.

### DIFF
--- a/packages/graphql-typescript-definitions/src/print/document/language.ts
+++ b/packages/graphql-typescript-definitions/src/print/document/language.ts
@@ -110,6 +110,10 @@ function tsTypeForObjectField(
   context: OperationContext,
 ) {
   const {inlineFragments = []} = field;
+  const possibleTypes =
+    isInterfaceType(graphQLType) || isUnionType(graphQLType)
+      ? context.ast.schema.getPossibleTypes(graphQLType)
+      : [];
 
   if (inlineFragments.length) {
     const fragmentTypes = [...inlineFragments].map((inlineFragment) =>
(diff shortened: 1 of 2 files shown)


loic-d:
cc/ @veenitmalik @denislaliberte for visibility

loic-d:
Could this PR break `web` for `union` using a single type?

lemonmade:
@loic-d there might be some breaking cases in Web, this is definitely a breaking change for the library.

PR #69, opened 2019-03-05
Allow for exporting multiple operations and fragments from the same file

alexkirsz:
`graphql-tag/loader` [allows for more than one exported operation type](https://github.com/apollographql/graphql-tag#support-for-multiple-operations), but `graphql-typescript-definitions` will only pick up the last operation defined. If an operation is defined, it will also ignore all fragments defined within the file.

This PR addresses that and allows for exporting multiple operations and fragments from the same file.

```ts
import { ProjectsQuery, ResultsQuery, ProjectFragmentData } from './Queries.gql';
```

~~Note that this is a **breaking** change: since the default export of a module is actually a composition of its definition and not necessarily a single operation, I've updated the type of the default export to `graphql`'s own untyped `DocumentNode` to better reflect this behavior. This means that existing code will break as typechecking will fail.~~

```ts
import ProjectsQuery from './Queries.gql';

// `ProjectsQuery` is an untyped graphql.DocumentNode that also contains
//  the definitions for `ResultsQuery`.
```

**(Implemented)** A backwards compatible but perhaps volatile change would be to keep the current behaviour when we detect that only one operation was defined within the file. This *might* break unexpectedly for existing files with multiple operations, but that would be indicative of an issue with the GraphQL definition file itself as only one operation was expected. Also, I don't know what happens when you feed a `DocumentNode` with more than one operation defined within to either Apollo or Relay, but the current behaviour might be deceptive in that regard.

**Disclaimer**: This PR fails to address the fact that `graphql-tag/loader` will also set every operation defined within the file as a property of the returned `DocumentNode`.

I'm putting this here as a heads up if you want to take a look at this change, as I will follow up with other contributions during the week, as well as take care of fixing/writing tests.

### DIFF
--- a/packages/graphql-typescript-definitions/src/index.ts
+++ b/packages/graphql-typescript-definitions/src/index.ts
@@ -59,7 +59,7 @@ export interface SchemaBuild {
 export interface DocumentBuild {
   documentPath: string;
   definitionPath: string;
-  operation?: Operation;
+  operations: Operation[];
   fragments: Fragment[];
 }
 
@@ -345,7 +345,7 @@ export class Builder extends EventEmitter {
     this.emit('build:docs', {
       documentPath: file.path,
       definitionPath,
-      operation: file.operation,
+      operations: file.operations,
       fragments: file.fragments,
     });
   }
@@ -434,7 +434,7 @@ function getSchemaTypesPath(
 
 interface File {
   path: string;
-  operation?: Operation;
+  operations: Operation[];
   fragments: Fragment[];
 }
 
@@ -451,7 +451,7 @@ function groupOperationsAndFragmentsByFile({operations, fragments}: AST) {
       if (!file) {
         file = {
           path: item.filePath,
-          operation: undefined,
+          operations: [],
           fragments: [],
         };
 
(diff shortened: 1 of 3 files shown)


alexkirsz:
I've updated this change to be as backward compatible as possible, as mentioned in the original comment. I've also updated the failing test cases and added new tests.

lemonmade:
Hmm, interesting, thanks for taking a look at this. Intellectually I knew this was possible but we just never do it at Shopify. So, just so I am clear, the behaviour of having multiple operations is that each ends up getting exported as effectively a named export (in pascal case?), and there is also a "default" export (`module.exports =`) that is... what exactly? I am not totally sure I understand that since the default export does definitely work, so I guess it is also a document node?

alexkirsz:
The default export is always a DocumentNode, which I believe contains all the types defined within the file.

lemonmade:
I think for now we are going to keep this restricted to only return a single operation. We don't use this feature in our apps, and our new [Webpack/ Jest transforms](https://github.com/Shopify/graphql-tools-web/tree/master/packages/graphql-mini-transforms) only output a single operation to avoid unnecessary runtime impact

alexkirsz:
@lemonmade Fair enough, but you might still want to throw whenever more than one operation/fragment is defined from within a single file since this is unsupported.

PR #70, opened 2019-03-06
Accept a config option

alexkirsz:
I'm using the Builder API directly, but my project does not define a `.graphqlconfig` file, which means that I must create a temporary one just for the sake of running the type generation operation.

This change adds the possibility to provide a `config` option to the `Builder` options interface in order to provide your own custom `GraphQLConfig` instance.


lemonmade:
This is shipped in v0.17 :)

PR #71, opened 2019-03-06
Remove contents of the ~ directory

alexkirsz:
I don't think this directory was supposed to be checked in version control?


lemonmade:
Whoops, thanks so much @alexkirsz!

PR #77, opened 2019-04-12
Update eslint-plugin-shopify

cartogram:
Updates eslint and eslint-plugin-shopify to their latest versions. I also updated the types for jest.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,15 +22,18 @@
   "eslintConfig": {
     "extends": [
       "plugin:shopify/typescript",
-      "plugin:shopify/typescript-prettier",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -22,15 +22,16 @@
   "eslintConfig": {
     "extends": [
       "plugin:shopify/typescript",
-      "plugin:shopify/typescript-prettier",
+      "plugin:shopify/react",
       "plugin:shopify/jest",
-      "plugin:shopify/node"
+      "plugin:shopify/react",
```
lemonmade: Duplicate of above

PR #78, opened 2019-04-18
[DO_NOT_MERGE] splitting node exports of graphql-tool-utilities into new module

patsissons:
In order to support usage of `graphql-tool-utilities` in a web context, we need to split the module into `node` and `default` exports. `node` will include everything in the `default` exports in addition to the `graphql-config` tooling, which requires a `node` context.

most imports will go on unaffected and can continue using `import {} from 'graphql-tool-utilities';`. However, for node specific usages, the import signature will change to `import {} from 'graphql-tool-utilities/node';`

### Reviewer Notes

Should we just create a new package for `graphql-config-utilities` instead of all these changes?

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 *.d.ts
+packages/graphql-tool-utilities/dist/**/*
 packages/graphql-typescript-definitions/test/fixtures/**/*.ts
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ node_modules
 packages/**/*.d.ts
 packages/graphql-typed/*.js
 packages/graphql-tool-utilities/*.js
+!packages/graphql-tool-utilities/dist/**/*
 packages/graphql-validate-fixtures/test/fixtures/**/schema.json
 packages/graphql-typescript-definitions/test/fixtures/**/schema.json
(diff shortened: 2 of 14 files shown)


patsissons:
@lemonmade that's how this started, but because monorepo it has to actually be a typescript module and that module needs to point to the src path at compile time and the output path at runtime. that's why i have a dummy `node.ts` at the root to solve compile time linking, and then `dist/*` files that get copied back to the root for runtime linking. Trust me when i say _this is ugly AF_ and i don't like it at all. Personally i would prefer to split out the `graphql-config` utils into its own package. If you have any alternative suggestions i'm very open to a new strategy haha 😁

lemonmade:
I am not sure I understand why monorepo matters here. It's just a file that points at the dist. If you need to create an additional mapping to the source in order to satisfy tests (which run against source), you can do so in the Jest config. We do this in Quilt for server-related code:

https://github.com/Shopify/quilt/blob/master/packages/react-effect/server.js
https://github.com/Shopify/quilt/blob/master/jest.config.js#L10-L13

lemonmade:
I wouldn't be opposed to separate package either 👍

patsissons:
I'll poke at a separate package then. it just makes more sense since to me.

patsissons:
replaced by #79

PR #79, opened 2019-04-18
splitting graphql-config utilities into new graphql-config-utilities

patsissons:
`graphql-config` uses file system libraries in `nodejs` but the AST utilities in `graphql-tool-utilities` can be used without `graphql-config` so it makes more sense to split the two pieces of functionality into separate packages so that they can be imported independently.

replaces #78


packages/graphql-config-utilities/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "graphql-config-utilities",
+  "description": "Common utilities for graphql-config",
+  "version": "0.0.0",
```
patsissons: i think this makes sense to start with right? so that lerna can bump to `0.1.0` on first publish?

packages/graphql-config-utilities/src/config.ts
```diff
@@ -1,14 +1,72 @@
+import {existsSync} from 'fs';
 import {promisify} from 'util';
 import {GraphQLConfig, GraphQLProjectConfig} from 'graphql-config';
 
-import './augmentations';
```
patsissons: a good time to 🔥 the augmentations

packages/graphql-config-utilities/src/config.ts
```diff
@@ -1,14 +1,72 @@
...
 
 // we need to use an import/require here because it does not force consumers to
 // enable esModuleInterop in tsconfig.json
 import glob = require('glob');
 
+export const defaultGraphQLProjectName = 'GraphQL';
+
+// temporary utility until `graphql-config` supports this function natively
+// see: https://github.com/prisma/graphql-config/pull/113
+export function resolvePathRelativeToConfig(
```
patsissons: i am merging the old `utilities.ts` into `config.ts` since they are essentially the same thing now.

packages/graphql-config-utilities/src/polyfills.ts
```diff
@@ -0,0 +1,16 @@
+// config.ts uses Array.flatmap
+import 'core-js/modules/es7.array.flat-map';
```
patsissons: this seems like the least invasive way to pull in individual functions, couldn't find actual docs on this though...

packages/graphql-config-utilities/src/polyfills.ts
```diff
@@ -0,0 +1,16 @@
...
+
+declare global {
+  interface Array<T> {
+    flatMap<U, This = undefined>(
```
patsissons: rather than pull in typings or the `esnext.array` lib, i'm just defining `flatMap` inline (copy paste from `typescript/lib/lib.es2019.array.d.ts`)

packages/graphql-tool-utilities/README.md
```diff
@@ -16,51 +16,7 @@ yarn add graphql-tool-utilities
 
 ## Usage
 
-### Configuration
```
patsissons: should i add a mention in here for where these functions now live?

packages/graphql-tool-utilities/package.json
```diff
@@ -17,7 +17,7 @@
   "bugs": {
     "url": "https://github.com/shopify/graphql-tools-web/issues"
   },
-  "homepage": "https://github.com/shopify/graphql-tools-web/blob/master/packages/graphql-typescript-definitions",
+  "homepage": "https://github.com/shopify/graphql-tools-web/blob/master/packages/graphql-tool-utilities",
```
patsissons: long outstanding issue i noticed 😄

packages/graphql-tool-utilities/src/polyfills.ts
```diff
@@ -1,3 +1,3 @@
 // apollo-codegen uses Object.entries and Array.flatmap
-import 'core-js/es7/object';
-import 'core-js/es7/array';
+import 'core-js/modules/es7.object.entries';
+import 'core-js/modules/es7.array.flat-map';
```
patsissons: might as well trim some fat here too while i'm at it

packages/graphql-tool-utilities/tsconfig.json
```diff
@@ -7,5 +7,5 @@
     "outDir": "lib"
   },
   "include": ["./**/*.ts"],
-  "exclude": ["./tests/**", "./lib/**", "node_modules/**"]
+  "exclude": ["./lib/**", "node_modules/**"]
```
patsissons: no more tests in this package... 🤷‍♂️

packages/graphql-config-utilities/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+This utility reads schema information from a [`.graphqlconfig`](https://github.com/prisma/graphql-config) file in the project root. The configuration can contain one nameless project or many named projects. The configuration is compatible with the [vscode-graphql extension](https://github.com/prisma/vscode-graphql). This extension provides syntax highlighting and autocomplete suggestions for graphql files.
+
+Each project specifies a `schemaPath`, `include`, and `exclude` globs. Glob patterns match paths relative to the location of the configuration file. Omit `exclude` if empty.
+
+See the [official specification documentation](https://github.com/prisma/graphql-config/blob/master/specification.md#use-cases) for more detail and examples.
+
+#### Examples
+
+A single nameless project configuration
```
lemonmade: A little weird that these aren't titles or sentences

packages/graphql-config-utilities/package.json
lemonmade: Yeah, but probably bump to 1.0 instead

packages/graphql-config-utilities/src/polyfills.ts
lemonmade: Is it our own code that uses flatMap? If so, can we just avoid it?

packages/graphql-tool-utilities/README.md
lemonmade: I don't think you need to worry about it

packages/graphql-tool-utilities/package.json
lemonmade: lol

packages/graphql-config-utilities/README.md
patsissons: yea good point, i'll convert them to titles

packages/graphql-config-utilities/src/polyfills.ts
patsissons: yea i can probably do that

PR #85, opened 2019-07-07
Bump apollo-codegen-core version

gavrix:
Bump apollo-codegen-core version.

#### Why
apollo-codegen-core we are currently using has some weird hard-coded handling of `@client` directives in graphql queries and therefore doesn't work for generating type definitions for queries with such directives.

I tested our code-generation script with newer version of `apollo-codegen-core` as dependency for `graphql-tool-utilities` and it went smoothly — so outdated `apollo-codegen-core` was the only problem.


gavrix:
Forgot to push that one too. Done 👍

PR #86, opened 2019-07-26
Update `eslint-plugin-shopify` to v30

cartogram:


packages/graphql-typescript-definitions/src/print/document/language.ts
```diff
@@ -281,7 +281,12 @@ function tsTypeForGraphQLType(
     : graphQLType;
 
   if (isScalarType(unwrappedGraphQLType)) {
-    if (scalarTypeMap.hasOwnProperty(unwrappedGraphQLType.name)) {
+    if (
+      Object.prototype.hasOwnProperty.call(
```
cartogram: https://eslint.org/docs/rules/no-prototype-builtins

PR #90, opened 2019-10-18
Allow custom scalars that alias built-in types

ryanw:
I have a custom type named `Seconds` in my GraphQL schema that is really just an `Int`. The current implementation of `graphql-typescript-definitions` doesn't allow this. So here's a PR that lets you alias any built-in JS type.

The `package` property in custom-scalars is now optional. If left out then no import will be performed, and the `name` is assumed to be a globally available type.

e.g.

    --custom-scalars '{"Seconds": {"name": "number"}}'

Will become:

    export type Seconds = number;


lemonmade:
Publish in 0.21.0

Issue #92, opened 2019-10-22
[graphql-mini-transforms] Unable to import fragment file in javascript


pleunv:
Hi guys! I'm looking into `graphql-mini-transforms` as possible replacement for `graph-tag/loader`, but ran into one annoying issue. We have some instances where we use apollo's [writeFragment/readFragment](https://www.apollographql.com/docs/react/caching/cache-interaction/#writequery-and-writefragment) to do optimistic store updates. In order to do this we import a `gql` file only containing a fragment definition from a javascript file, and pass that along to `writeFragment({ fragment })`/`readFragment({ fragment })`.

Trying this with `graphql-mini-transforms` unfortunately throws the following error:
```js
ERROR in ./src/.../OptimisticCommentFragment.gql
Module build failed (from ./node_modules/graphql-mini-transforms/webpack.js):
GraphQLError: Syntax Error: Unexpected <EOF>
    at syntaxError (/Users/.../node_modules/graphql/error/syntaxError.js:15:10)
    at Parser.unexpected (/Users/.../node_modules/graphql/language/parser.js:1463:41)
    at Parser.parseDefinition (/Users/.../node_modules/graphql/language/parser.js:157:16)
    at Parser.many (/Users/.../node_modules/graphql/language/parser.js:1518:26)
    at Parser.parseDocument (/Users/.../node_modules/graphql/language/parser.js:111:25)
    at Object.parse (/Users/.../node_modules/graphql/language/parser.js:36:17)
    at Object.cleanDocument (/Users/.../node_modules/graphql-mini-transforms/lib/document.js:15:42)
    at Object.<anonymous> (/Users/.../node_modules/graphql-mini-transforms/lib/webpack.js:23:68)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/.../node_modules/graphql-mini-transforms/lib/webpack.js:4:58)
```
Any idea what's causing this? Or do fragments currently only work inside of other `gql` files, with the `#import` syntax?

lemonmade:
Hmm yeah the current implementation strips fragments that aren’t used for anything; let me try to put something together to properly export it if there's only a single fragment in the file (I don’t think we’ll provide support for multiple fragments in the file to be consistent with how query/ mutation are done). `writeFragment` and `readFragment` are definitely a good use case for producing the AST.

ajhyndman:
Hey!  I also just ran into this issue, while trying out `graphql-mini-transforms`.

[...]

PR #95, opened 2019-11-15
Export fragment as default when it’s the only export

lemonmade:
Fixes https://github.com/Shopify/graphql-tools-web/issues/92


lemonmade:
Going to close this for now. There are some issues with this approach, most notably that it does not work with imported fragments. Anyone is free to pick this up, but I won't have time to address this in the next few months unfortunately.

PR #99, opened 2020-02-15
add generateId option to graphql-mini-transforms

dsanders11:
Effectively the same as the `generateId` option from `graphql-persisted-document-loader`.

I've got a use-case where I'd like to make the ID a signature of the hash, rather than just the hash, so that the server can trust queries presented by the client, since the IDs were signed at build-time. Removes the need to distribute a generated mapping like that from `@shopify/webpack-persisted-graphql-plugin`.

Being able to customize the generated ID would make this possible, and it's a pretty small change.

### DIFF
--- a/packages/graphql-mini-transforms/README.md
+++ b/packages/graphql-mini-transforms/README.md
@@ -61,7 +61,7 @@ fragment ProductVariantId on ProductVariant {
 
 #### Options
 
-##### simple
+##### `simple`
 
 This option changes the shape of the value exported from `.graphql` files. By default, a `graphql-typed` `DocumentNode` is exported, but when `simple` is set to `true`, a `SimpleDocument` is exported instead. This representation of GraphQL documents is smaller than a full `DocumentNode`, but generally won’t work with normalized GraphQL caches.
 
@@ -82,11 +82,9 @@ module.exports = {
 
 If this option is set to `true`, you should also use the `jest-simple` transformer for Jest, and the `--export-format simple` flag for `graphql-typescript-definitions`.
 
-##### generateId
+##### `generateId`
 
-This option changes the identifier value used. By default the hash of the minified GraphQL document is used as the identifier value, but when `generateId` is provided
-the return value is used as the identifier value. `generateId` should be a function which takes a single parameter, the normalized GraphQL document source as a string,
-and it should return a string value.
+This option changes the identifier value used. By default the hash of the minified GraphQL document is used as the identifier value, but when `generateId` is provided the return value is used as the identifier value. `generateId` should be a function which takes a single parameter, the normalized GraphQL document source as a string, and it should return a string value.
 
 ```js
 module.exports = {
--- a/packages/graphql-mini-transforms/README.md
+++ b/packages/graphql-mini-transforms/README.md
@@ -61,7 +61,9 @@ fragment ProductVariantId on ProductVariant {
 
 #### Options
 
-This loader accepts a single option, `simple`. This option changes the shape of the value exported from `.graphql` files. By default, a `graphql-typed` `DocumentNode` is exported, but when `simple` is set to `true`, a `SimpleDocument` is exported instead. This representation of GraphQL documents is smaller than a full `DocumentNode`, but generally won’t work with normalized GraphQL caches.
+##### simple
+
+This option changes the shape of the value exported from `.graphql` files. By default, a `graphql-typed` `DocumentNode` is exported, but when `simple` is set to `true`, a `SimpleDocument` is exported instead. This representation of GraphQL documents is smaller than a full `DocumentNode`, but generally won’t work with normalized GraphQL caches.
 
 ```js
 module.exports = {
@@ -79,6 +81,27 @@ module.exports = {
 ```
 
 If this option is set to `true`, you should also use the `jest-simple` transformer for Jest, and the `--export-format simple` flag for `graphql-typescript-definitions`.
+
+##### generateId
+
+This option changes the identifier value used. By default the hash of the minified GraphQL document is used as the identifier value, but when `generateId` is provided
+the return value is used as the identifier value. `generateId` should be a function which takes a single parameter, the normalized GraphQL document source as a string,
+and it should return a string value.
+
+```js
+module.exports = {
+  module: {
+    rules: [
+      {
+        test: /\.(graphql|gql)$/,
+        use: 'graphql-mini-transforms/webpack',
+        exclude: /node_modules/,
+        options: {generateId: normalizedSource => someHash(normalizedSource)},
+      },
+    ],
+  },
+};
+```
 
 ### Jest
 
--- a/packages/graphql-mini-transforms/README.md
+++ b/packages/graphql-mini-transforms/README.md
@@ -84,7 +84,7 @@ If this option is set to `true`, you should also use the `jest-simple` transform
 
 ##### `generateId`
 
-This option changes the identifier value used. By default the hash of the minified GraphQL document is used as the identifier value, but when `generateId` is provided the return value is used as the identifier value. `generateId` should be a function which takes a single parameter, the normalized GraphQL document source as a string, and it should return a string value.
+This option changes the identifier value used. By default the hash of the minified GraphQL document is used as the identifier value, but when `generateId` is provided the return value is used as the identifier value. `generateId` should be a function which takes a single parameter, an object with the normalized GraphQL document source as a string under the `source` key, and it should return a string value.
 
 ```js
 module.exports = {
@@ -94,7 +94,7 @@ module.exports = {
         test: /\.(graphql|gql)$/,
         use: 'graphql-mini-transforms/webpack',
         exclude: /node_modules/,
-        options: {generateId: normalizedSource => someHash(normalizedSource)},
+        options: {generateId: ({source}) => someHash(source)},
       },
     ],
   },
--- a/packages/graphql-mini-transforms/src/document.ts
+++ b/packages/graphql-mini-transforms/src/document.ts
@@ -16,16 +16,20 @@ import {DocumentNode, SimpleDocument} from 'graphql-typed';
 const IMPORT_REGEX = /^#import\s+['"]([^'"]*)['"];?[\s\n]*/gm;
 const DEFAULT_NAME = 'Operation';
 
-function defaultGenerateId(normalizedSource: string) {
+export interface GenerateIdArguments {
+  source: string;
+}
+
+function defaultGenerateId({source}: GenerateIdArguments) {
   // This ID is a hash of the full file contents that are part of the document,
   // including other documents that are injected in, but excluding any unused
   // fragments. This is useful for things like persisted queries.
-  return createHash('sha256').update(normalizedSource).digest('hex');
+  return createHash('sha256').update(source).digest('hex');
 }
 
 export interface CleanDocumentOptions {
   removeUnused?: boolean;
-  generateId?: (normalizedSource: string) => string;
+  generateId?: (args: GenerateIdArguments) => string;
 }
 
 export function cleanDocument(
@@ -50,8 +54,8 @@ export function cleanDocument(
 
   const id =
     generateId === undefined
-      ? defaultGenerateId(normalizedSource)
-      : generateId(documentSource);
+      ? defaultGenerateId({source: normalizedSource})
+      : generateId({source: documentSource});
 
   Reflect.defineProperty(normalizedDocument, 'id', {
     value: id,
--- a/packages/graphql-mini-transforms/src/document.ts
+++ b/packages/graphql-mini-transforms/src/document.ts
@@ -16,9 +16,26 @@ import {DocumentNode, SimpleDocument} from 'graphql-typed';
 const IMPORT_REGEX = /^#import\s+['"]([^'"]*)['"];?[\s\n]*/gm;
 const DEFAULT_NAME = 'Operation';
 
+function defaultGenerateId(normalizedSource: string) {
+  // This ID is a hash of the full file contents that are part of the document,
+  // including other documents that are injected in, but excluding any unused
+  // fragments. This is useful for things like persisted queries.
+  return createHash('sha256')
+    .update(minifySource(normalizedSource))
+    .digest('hex');
+}
+
+export interface CleanDocumentOptions {
+  removeUnused?: boolean;
+  generateId?: (normalizedSource: string) => string;
+}
+
 export function cleanDocument(
   document: UntypedDocumentNode,
-  {removeUnused = true} = {},
+  {
+    removeUnused = true,
+    generateId = defaultGenerateId,
+  }: CleanDocumentOptions = {},
 ): DocumentNode<any, any, any> {
   if (removeUnused) {
     removeUnusedDefinitions(document);
@@ -35,10 +52,7 @@ export function cleanDocument(
     stripLoc(definition);
   }
 
-  // This ID is a hash of the full file contents that are part of the document,
-  // including other documents that are injected in, but excluding any unused
-  // fragments. This is useful for things like persisted queries.
-  const id = createHash('sha256').update(normalizedSource).digest('hex');
+  const id = generateId(print(document));
 
   Reflect.defineProperty(normalizedDocument, 'id', {
     value: id,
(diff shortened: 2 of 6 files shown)


packages/graphql-mini-transforms/src/document.ts
```diff
@@ -33,12 +50,7 @@ export function cleanDocument(
     stripLoc(definition);
   }
 
-  // This ID is a hash of the full file contents that are part of the document,
-  // including other documents that are injected in, but excluding any unused
-  // fragments. This is useful for things like persisted queries.
-  const id = createHash('sha256')
-    .update(normalizedSource)
-    .digest('hex');
+  const id = generateId(print(document));
```
lemonmade, line 53: I think we'd need this to continue referencing `normalizedSource` (which is printed + minified).

packages/graphql-mini-transforms/src/webpack.ts
```diff
@@ -3,14 +3,22 @@ import {dirname} from 'path';
...
 ) {
   this.cacheable();
 
+  const options = this.query;
+
+  if (options && typeof options !== 'object') {
+    throw new Error(
+      '@shopify/graphql-mini-transforms only supports options as an object',
+    );
+  }
```
lemonmade, line 20: I think the common pattern for webpack loaders it to use the loader-utils package to normalize the options.

packages/graphql-mini-transforms/src/webpack.ts
```diff
@@ -20,8 +28,19 @@ export default async function graphQLLoader(
   }
 
   try {
+    const cleanDocumentOptions = {} as CleanDocumentOptions;
+
+    if (options) {
+      cleanDocumentOptions.generateId = options.generateId;
+    }
```
lemonmade, line 35: We generally use ternaries over mutating objects `const cleanDocumentOptions = {generateId: options ? option.generateId, undefined}`

lemonmade (changes requested):
Great feature, very smart to put the signing at build-time so you can treat these as trusted 👍 just a few nits but then I'll be happy to merge this in. Sorry it took so long to get to this!

dsanders11:
@lemonmade, thanks for taking a look. I'll try to get this PR updated with your requested changes when I get some time. In the meanwhile, I'm going to respond to any comments which need clarification.

packages/graphql-mini-transforms/src/document.ts
dsanders11: The minifying is still done for the default implementation, it was just moved to `defaultGenerateId`. This allows the user-provided implementation of `generateId` to receive the non-minified version.

It was implemented this way as a result of testing my implementation with the signing. Since the server in that case is fed the queries by the client, and checks the signature to confirm they're legit, the query sent to the server needs to be identical to what was hashed and signed at build time. I found that the query which is sent to the server is the non-minified version, so if the signature was done on the minified version the signature wouldn't match the query payload. So, the unminified version of the query is provided to `generatedId`. It might be useful to export `minifySource` if a user may require it for implementing `generateId`, but I didn't think it necessary at the time.

Note this is from memory, and I haven't looked at this in months, so bear with me if I missed something.

dsanders11:
@lemonmade, this has now been updated to be based on the latest `main` and address your comments.

dsanders11:
Rebased onto latest `main`.

@lemonmade @alexandcote think you could take a look at this PR? Would be nice to get it merged so it doesn't keep slipping into a conflicted state.

alexandcote (commented):
Hi @dsanders11 

First of all, thanks a lot for your contribution. I reviewed your PR, nothing major only nit 💅.
Now that I'm following this PR, feel free to ping me again when you have time for the small tweaks. If you know you will not have time to do the small changes, let me know I can probably take time before the holiday.

Again, thanks a lot for your contribution and sorry for the long delay on this one 😥 ...

packages/graphql-mini-transforms/src/webpack.ts
```diff
@@ -3,33 +3,59 @@ import {dirname} from 'path';
...
 
   if (done == null) {
     throw new Error(
       '@shopify/graphql-loader does not support synchronous processing',
     );
   }
 
+  const cleanDocumentOptions = {
+    generateId: options.generateId,
+  } as CleanDocumentOptions;
```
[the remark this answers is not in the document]
dsanders11: I think you're right. I wasn't very familiar with TypeScript at the time of this PR, so there's bound to be stuff that's not as clean as it could be. 🙂 Using the suggested change.

alexandcote (approved):
This PR looks good to me ✅ . I will follow with @lemonmade tomorrow. Thanks for your quick rework 💯

dsanders11:
@alexandcote, I think I snuck in one more commit while you were making that comment, fad6a450ae6ef2eed98a80389324517d032e718b. Per your comment about `print` being done twice, I also noticed `minifySource` would be done twice for the default generate ID, so I reworked it to inline `defaultGenerateId` so that it could use the already minified source instead of doing it again.

So that's just a little further optimization, but I'm also fine to drop that commit. Just thought I'd point it out.

Thanks for taking a look and the quick feedback.

packages/graphql-mini-transforms/src/document.ts
```diff
@@ -28,17 +33,26 @@ export function cleanDocument(
...
-  // including other documents that are injected in, but excluding any unused
-  // fragments. This is useful for things like persisted queries.
-  const id = createHash('sha256').update(normalizedSource).digest('hex');
+  let id: string;
+
+  if (generateId === undefined) {
+    // This ID is a hash of the full file contents that are part of the document,
+    // including other documents that are injected in, but excluding any unused
+    // fragments. This is useful for things like persisted queries.
+    id = createHash('sha256')
```
alexandcote, line 50: I liked that the generator was extracted out.

Suggestion here: 

```ts
function defaultGenerateId(normalizedSource: string) {	
  // This ID is a hash of the full file contents that are part of the document,	
  // including other documents that are injected in, but excluding any unused	
  // fragments. This is useful for things like persisted queries.	
  return createHash('sha256')	
    .update(normalizedSource)	
    .digest('hex');	
}
...

// (Don't add those comments, they are only for the review)
// documentSource is not minified
// normalizedSource is already minified
const id = generateId == null ? defaultGenerateId(normalizedSource) : generateId(documentSource);
```

The current commit has an issue because you still call `minifySource` twice 😄 
```

**Review by alexandcote — CHANGES_REQUESTED** [review]

File: packages/graphql-mini-transforms/src/document.ts (hunk 5 above)
Comment (dsanders11) [reply packages/graphql-mini-transforms/src/document.ts:50]: > The current commit has an issue because you still call `minifySource` twice 😄

Ha! That's what I get for rushing.

> I liked that the generator was extracted out.

I did too. I don't love the argument being different for the two, since that might cause confusion, if someone looks at `defaultGenerateId` and doesn't realize the argument is slightly different than user-provided functions. However, it's not a major concern, so I'll go ahead and implement your suggestion.

**Comment by dsanders11** [comment]
Alright, changes made, I'll stop touching it so it can be merged. 🙂

**Review by alexandcote — APPROVED** [review]
💯

**Review by michenly — APPROVED** [review]
@alexandcote had asked me to take a look. And it all looks great!

**Comment by alexandcote** [comment]
I re-request a review from @lemonmade. Probably this will have to wait until after the holiday tho.

File: packages/graphql-mini-transforms/README.md [hunk 6]
```diff
@@ -61,7 +61,9 @@ fragment ProductVariantId on ProductVariant {
 
 #### Options
 
-This loader accepts a single option, `simple`. This option changes the shape of the value exported from `.graphql` files. By default, a `graphql-typed` `DocumentNode` is exported, but when `simple` is set to `true`, a `SimpleDocument` is exported instead. This representation of GraphQL documents is smaller than a full `DocumentNode`, but generally won’t work with normalized GraphQL caches.
+##### simple
```
lemonmade, line 64: Can we put this heading and the `generateId` one in backticks?

packages/graphql-mini-transforms/README.md
```diff
@@ -80,6 +82,27 @@ module.exports = {
 
 If this option is set to `true`, you should also use the `jest-simple` transformer for Jest, and the `--export-format simple` flag for `graphql-typescript-definitions`.
 
+##### generateId
+
+This option changes the identifier value used. By default the hash of the minified GraphQL document is used as the identifier value, but when `generateId` is provided
+the return value is used as the identifier value. `generateId` should be a function which takes a single parameter, the normalized GraphQL document source as a string,
```
lemonmade, line 88: I think we generally do soft wraps instead of hard wraps

packages/graphql-mini-transforms/src/document.ts
```diff
@@ -28,17 +40,18 @@ export function cleanDocument(
...
   for (const definition of normalizedDocument.definitions) {
     stripLoc(definition);
   }
 
-  // This ID is a hash of the full file contents that are part of the document,
-  // including other documents that are injected in, but excluding any unused
-  // fragments. This is useful for things like persisted queries.
-  const id = createHash('sha256').update(normalizedSource).digest('hex');
+  const id =
+    generateId === undefined
```
lemonmade, line 52: Instead of checking this here, can you add a default value when destructuring the argument?

packages/graphql-mini-transforms/tests/webpack.test.ts
```diff
@@ -70,6 +70,14 @@ describe('graphql-mini-transforms/webpack', () => {
     );
   });
 
+  it('has option for custom ID generate function', async () => {
```
lemonmade, line 73:
```suggestion
  it('uses the generateId option for the document ID when provided', async () => {
```

packages/graphql-mini-transforms/tests/webpack.test.ts
```diff
@@ -70,6 +70,14 @@ describe('graphql-mini-transforms/webpack', () => {
...
+    const result = await extractDocumentExport(
+      `query Shop { shop { id } }`,
+      createLoaderContext({query: {generateId: () => 'foo'}}),
+    );
+    expect(result).toHaveProperty('id', 'foo');
```
lemonmade, line 78: It would be nice to have the `generateId` option be a spy so we can check that it was called with what we expect (or do this in a separate function)

packages/graphql-mini-transforms/src/webpack.ts
```diff
@@ -3,21 +3,37 @@ import {dirname} from 'path';
 import {loader} from 'webpack';
 import {parse, DocumentNode} from 'graphql';
 import {getOptions} from 'loader-utils';
+import validateOptions from 'schema-utils';
 
 import {cleanDocument, extractImports, toSimpleDocument} from './document';
 
 interface Options {
+  generateId?: (normalizedSource: string) => string;
```
lemonmade, line 11: Instead of this just accepting the source as a string, can we make the argument be an object of `{source: string}`? Right now source is the only argument we need to pass in, but this signature gives us more flexibility to add things like the path, or more information about the concatenated fragments, if that is needed later on.

lemonmade (changes requested):
Feature makes sense to me, code overall looks great, just some minor suggestions and nits from me.

packages/graphql-mini-transforms/src/document.ts
dsanders11: I'm not sure I follow this remark. What default value are you suggesting? The line you've commented on would still be required, so do you just want to explicitly destructure to `undefined` or `null`?

dsanders11 (commented):
Added a comment

packages/graphql-mini-transforms/src/document.ts
alexandcote: I think what he proposed is something like : 
```ts
{removeUnused = true, generateId = defaultGenerateId}: CleanDocumentOptions = {},
...
const id = generateId(normalizedSource)
```

If we want to do something like this, both `defaultGenerateId` and `generateId` will have to receive the same source. Currently, `defaultGenerateId` receives the normalized source and the `generateId` receives the document source.

I don't remember why `generateId` needs the document source.

dsanders11: Well, that's what the code originally looked like... the commits and comments leading to that change are still there for review. That's why I'm a bit confused.

While changing the code to prevent printing the source twice, I also changed it so that `minifySource` was only called once since the extra operations seemed to be a concern.

The general `generateId` needs the unminifed source (this is explained in a previous comment). The default implementation minifies it, so it was happening twice. That motivated that change.

I can change it back, but I'm going to address the other comments first and leave this for now, so I don't go in circles on this.

packages/graphql-mini-transforms/src/webpack.ts
dsanders11: Changed in 13b8012b1b91f340be0491cf5c541dc275d3e349.

packages/graphql-mini-transforms/tests/webpack.test.ts
dsanders11: Done in c8c54ff. It got a bit mucky because `stripIndent` strips trailing whitespace, but the actual value has a trailing newline. Originally just went to concatenate a newline on after `stripIndent`, but ESLint didn't like that (and I didn't much like disabling the warning for that line), so settled on `expect.toStartWith` to skip around the trailing newline issue. Suggestions for better solution welcome.

packages/graphql-mini-transforms/src/document.ts
lemonmade: Sorry if I just missed this, but I am a bit confused why `generateId` needs to take the "full" document source, not the minified version. I assumed everything should be operating on the minified source, since that includes all the relevant fragments and normalizes away differences in whitespace and commas.

packages/graphql-mini-transforms/tests/webpack.test.ts
```diff
@@ -70,6 +71,32 @@ describe('graphql-mini-transforms/webpack', () => {
     );
   });
 
+  it('uses the generateId option for the document ID when provided', async () => {
+    // Mock function can't be called directly due to API schema validation, see
+    // https://github.com/facebook/jest/issues/6329
```
lemonmade, line 76: Annoying :(

lemonmade (changes requested):

packages/graphql-mini-transforms/src/document.ts
alexandcote: I think we need to take a decision on what we want here. Do we want to pass the minified version or the "full" document source to the `generateId` function? 

- Passing the minified version to the `generateId` would be the simplest solution but I think this was causing an issue for your use case @dsanders11 ?
- Passing the "full" document to both functions will result in minifying twice the document (not sure of the cost of this)
- Having the current implementation solves @dsanders11 limitation and ensures we only minify the source once.

The keypoint here is what is the limitation with the minified version @dsanders11 and does the cost of the modification is high.

dsanders11: @alexandcote, I think you've accurately summarized where things stand.

> * Passing the minified version to the generateId would be the simplest solution but I think this was causing an issue for your use case @dsanders11 ?

In no way trying to be rude, but all of the previous discussion and explanation still exists in this PR, nothing has been deleted. If you don't remember something about certain choices I'd suggest looking back on the other comments here for clarification. [My comment from September](https://github.com/Shopify/graphql-tools-web/pull/99#discussion_r495517313) explains why the non-minified version is needed for `generateId`. If that explanation isn't clear, I can try to clarify any point of confusion. The TL;DR is simply that the string given to `generateId` should match the query sent to the server, and the query sent to the server is the non-minified version.

lemonmade: Sorry, this is an unfortunate side effect of me/ us having let a PR review go for 4 months :p

I see where you are coming from on this, but I worry that the approach is very specific to one very special case where the printed, "normalized" document is identical to the original GraphQL source. This only happens when you aren't using fragments from external files, and you aren't letting the library automatically add `__typename` fields. IMO, this case is specific and uncommon enough (`__typename` additions are on by default), that we should push the pain of dealing with it to consumers' application code. An application can fairly easily run the same minification on the server source before computing the SHA (happy to export a utility for it if necessary), and that will work regardless of the fragment/ typename situation (and also works if you ever switch to our "simple" document format that doesn't need to be parsed and re-printed by the GraphQL client).

How do you feel about that?

lemonmade: If we don't do that, I think we need to give the "generate me an ID" option three different sources that might be relevant for them: the minified source code, the original document source code (no fragments or typenames embedded), and the "normalized" document that includes all the fragments and typenames.

michenly:
👋 So sorry for the trouble, we have move this package to https://github.com/Shopify/quilt/tree/main/packages/graphql-mini-transforms please move this PR over to quilt if you would still like to contribute 🙇‍♀️

PR #102, opened 2020-03-12
Fix random sizing of lists

shermam:
The way this was implemented before we would get only two possible sizes for the random list.
For example if I were to use list with `list([2,6])`, where we do `size[Math.round(Math.random())]` the only two possible indexes are 0 and 1, so `finalSize` in this case would be either 2 or 6, but would never be anything in between. So sometimes we would have an array of size 2, sometimes of size 6, but nothing else.

In my proposed implementation I take a random number from 0 to the difference between min and max by multiplying `Math.random` with that difference and add the result to the min value. This would give us a random number between the two values as advertised in the docs:
> an array randomly sized in that range

I hope this helps.

Regards.


shermam:
Oh, absolutely, I will try to get to that as soon as I can.

Thanks for quick response.

PR #105, opened 2020-04-09
upgrade fs-extra to 9

vsumner:
This PR upgrades `fs-extra` from `v6` to `v9`.

[Changelog](https://github.com/jprichardson/node-fs-extra/blob/master/CHANGELOG.md)
- Requires Node.js version 10 or greater 
- outputJson now outputs objects as they were when the function was called, even if they are mutated later 
- Cannot pass null as an options parameter to *Json* methods
- Peformance optimizations
- Refine copy*() handling of symlinks to properly detect symlinks that point to the same file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,6 @@
     "typescript": "3.4.1"
   },
   "dependencies": {
-    "fs-extra": "^6.0.1"
+    "fs-extra": "^9.0.0"
   }
 }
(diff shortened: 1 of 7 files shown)


alexandcote:
I think if we drop the support of node 8 we can 🔥 fs-extra from the dependencies.

The only reason why we use fs-extra is because of `mkdirp`. Since Node >= 10, we can use 
```
fs.mkdirSync(dirpath, { recursive: true })
```

alexandcote (approved):

PR #107, opened 2020-04-13
Upgrade yargs to 15 and jest to 25

vsumner:
This PR upgrades:

- `yargs` and `@types/yargs` to version `15` from `11`
- `jest` from `24` to `25`

See [Yargs Changelog](https://github.com/yargs/yargs/blob/master/CHANGELOG.md)
[Jest Changelog](https://github.com/facebook/jest/blob/master/CHANGELOG.md)

### DIFF
--- a/packages/graphql-typescript-definitions/package.json
+++ b/packages/graphql-typescript-definitions/package.json
@@ -35,10 +35,10 @@
     "@babel/types": "^7.4.0",
     "@types/babel__generator": "^7.0.2",
     "@types/chokidar": "^2.1.3",
-    "@types/yargs": "11.0.0",
     "chalk": "^4.0.0",
     "change-case": "^4.1.1",
     "chokidar": "^3.3.1",
+    "@types/yargs": "15.0.4",
     "fs-extra": "^9.0.0",
     "glob": "^7.1.2",
     "graphql": ">=14.5.0 <15.0.0",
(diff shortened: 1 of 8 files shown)


packages/graphql-typescript-definitions/package.json
```diff
@@ -25,7 +25,6 @@
     "prepublishOnly": "yarn build"
   },
   "devDependencies": {
-    "@types/chalk": "^2.2.0",
```
alexandcote, line 28: We don't need `@types/chalk` finally? 😃

packages/graphql-typescript-definitions/src/cli.ts
```diff
@@ -55,12 +54,16 @@ const argv = yargs
...
 
+type Args = typeof argv;
+
 const builder = new Builder({
   cwd: argv.cwd,
-  schemaTypesPath: argv.schemaTypesPath,
-  addTypename: argv.addTypename,
-  enumFormat: argv.enumFormat,
-  customScalars: normalizeCustomScalars(argv.customScalars),
+  schemaTypesPath: argv.schemaTypesPath as Args['schema-types-path'],
```
alexandcote, line 61: I think I prefer doing something like this instead.
```
const builder = new Builder({
  cwd: argv.cwd,
  schemaTypesPath: argv['schema-types-path'],
  addTypename: argv['add-typename'],
  enumFormat: argv['enum-format'],
  customScalars: normalizeCustomScalars(
    argv['custom-scalars'],
  ),
});
```

Any strong opinion?

alexandcote (approved):

packages/graphql-typescript-definitions/package.json
vsumner: It's a stub now. Chalk includes its own types.

packages/graphql-typescript-definitions/src/cli.ts
vsumner: Yeah that's better.

PR #114, opened 2020-04-27
"Simple" document types

lemonmade:
This PR introduces a new format that a GraphQL document can be exported in. This format, which I've called `simple` (open to other names; also considered `raw`/ `no-ast`), does not contain the full AST representation of the query (so, no longer a `DocumentNode`). Instead, it has the minimal amount of information needed by a GraphQL client that does a query-based cache: `id` (a hash of the minified document source; useful as a cache key), `source` (the minified document source; need to send as part of network requests), and `name` (the first operation name in the source; need to send as part of network requests). This representation is significantly smaller than the full AST.

While I was here, I also changed the webpack output of a `.graphql` file to be the result of `JSON.parse`ing a string, rather than embedding the object structure in the JavaScript file directly, as this can be [significantly faster for large objects](https://v8.dev/blog/cost-of-javascript-2019#json).

### DIFF
--- a/packages/graphql-mini-transforms/src/document.ts
+++ b/packages/graphql-mini-transforms/src/document.ts
@@ -3,21 +3,23 @@ import {createHash} from 'crypto';
 import {
   print,
   parse,
-  DocumentNode,
+  DocumentNode as UntypedDocumentNode,
   DefinitionNode,
   SelectionSetNode,
   ExecutableDefinitionNode,
+  OperationDefinitionNode,
   SelectionNode,
   Location,
 } from 'graphql';
+import {DocumentNode, SimpleDocument} from 'graphql-typed';
 
 const IMPORT_REGEX = /^#import\s+['"]([^'"]*)['"];?[\s\n]*/gm;
 const DEFAULT_NAME = 'Operation';
 
 export function cleanDocument(
-  document: DocumentNode,
+  document: UntypedDocumentNode,
   {removeUnused = true} = {},
-) {
+): DocumentNode<any, any, any> {
   if (removeUnused) {
     removeUnusedDefinitions(document);
   }
@@ -52,7 +54,7 @@ export function cleanDocument(
     configurable: false,
   });
 
-  return normalizedDocument;
+  return normalizedDocument as any;
 }
 
 export function extractImports(rawSource: string) {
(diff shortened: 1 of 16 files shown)


packages/graphql-mini-transforms/src/document.ts
```diff
@@ -66,7 +68,24 @@ export function extractImports(rawSource: string) {
   return {imports: [...imports], source};
 }
 
-function removeUnusedDefinitions(document: DocumentNode) {
+export function toSimpleDocument<Data, Variables, DeepPartial>(
+  document: DocumentNode<Data, Variables, DeepPartial>,
+): SimpleDocument<Data, Variables, DeepPartial> {
```
TheMallen, line 73: I guess this quietly changes the type for consumers not using simple as well? We should make sure to call that out in CHANGELOG as well

yarn.lock
```diff
@@ -1391,6 +1394,11 @@ bcrypt-pbkdf@^1.0.0:
   dependencies:
     tweetnacl "^0.14.3"
 
+big.js@^5.2.2:
```
TheMallen, line 1397: big

yarn.lock
```diff
@@ -779,10 +782,10 @@
     "@types/source-list-map" "*"
     source-map "^0.6.1"
 
-"@types/webpack@^4.41.11":
-  version "4.41.11"
-  resolved "https://registry.yarnpkg.com/@types/webpack/-/webpack-4.41.11.tgz#7b7f725397d3b630bede05415d34e9ff30d9771f"
-  integrity sha512-PtEZISfBMWL05qOpZN19hztZPt0rPuGQh5sbBP3bB4RrJgzdb0SScn47hdcMaoN1IgaU7NZWeDO6reFcKTK2iQ==
+"@types/webpack@*", "@types/webpack@^4.41.11":
```
TheMallen, line 785: nice dedupes

packages/graphql-mini-transforms/src/document.ts
lemonmade: I think it is only conditionally exported in this shape (see `jest`/ `webpack` sibling files, which consume these utilities)

packages/graphql-mini-transforms/src/document.ts
```diff
@@ -66,7 +68,24 @@ export function extractImports(rawSource: string) {
...
+  return {
+    id: document.id,
```
GoodForOneFare, line 75: Would `source` be an adequate cache key?

lemonmade: It is adequate but inefficient, especially for some queries that need to query enormous parts of a graph (like checkout's proposal query, or many of the online store team's queries for getting theme details).

PR #116, opened 2020-05-06
Fix chalk import to work with esModuleInterop

vsumner:
I introduced a change that incorrectly imported chalk. See https://github.com/chalk/chalk/issues/399
This PR fixes it.

### DIFF
--- a/packages/graphql-typescript-definitions/src/cli.ts
+++ b/packages/graphql-typescript-definitions/src/cli.ts
@@ -1,6 +1,6 @@
 /* eslint no-console: off */
 
-import * as chalk from 'chalk';
+import chalk from 'chalk';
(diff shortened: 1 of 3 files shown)


maciekglowacki:
I don't know if this is the right place to ask but on my project using chalk in version 4.1.0 import is still not working. 

Should I provide you with some additional data?

vsumner:
> I don't know if this is the right place to ask but on my project using chalk in version 4.1.0 import is still not working.
> 
> Should I provide you with some additional data?

Please do. Feel free to open an issue with any data you can provide.

tleunen:
Did you force update `graphql-validate-fixtures` to 0.12.1? I believe SK still use 0.12.0 so you might need to force the update of that package

